PR #121, opened 2015-03-20
BZ-1203696: application of proper use of Roles and Groups

porcelli:


manstis:
http://github.com/droolsjbpm/guvnor/compare/77b740f...d59abcf

PR #122, opened 2015-04-08
BZ-1206627: adapted to latest uberfire shutdown mechanism

porcelli:


manstis:
http://github.com/droolsjbpm/guvnor/commit/5959e828c

PR #123, opened 2015-04-24
Gwt upgrade n patter fly

porcelli:


manstis:
http://github.com/droolsjbpm/guvnor/compare/c7cb6e2...6c10631

guvnor-webapp doesn't compile due to a GWT module name relating to DashBuilder... when the DashBuilder PR is merged (David is doing that) I will revisit guvnor-webapp to check it works.

PR #129, opened 2015-05-20
BPMSPL-160: DRL text editor improvements.

manstis:


porcelli:
http://github.com/droolsjbpm/guvnor/commit/aeb6e5da9

PR #130, opened 2015-05-23
BPMSPL-74 -- adding operation to retrieve a (k)jar from the repository

mrietveld:
One of the requested features (aiming for milestone 1) is a REST operation to retrieve a process image.

However, the default way to do this is retrieve the process image .svg file that has been stored in the deployed deployment in the repository. This PR relates to retrieving a (k)jar from the repository on the server side in order to then extract the process image SVG file from the jar.

### DIFF
--- a/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
+++ b/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
@@ -40,6 +40,7 @@ import java.util.zip.ZipException;
 import java.util.zip.ZipFile;
 import java.util.zip.ZipInputStream;
 import java.util.zip.ZipOutputStream;
+
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 
@@ -65,6 +66,9 @@ import org.eclipse.aether.installation.InstallRequest;
 import org.eclipse.aether.installation.InstallationException;
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.repository.RepositoryPolicy;
+import org.eclipse.aether.resolution.ArtifactRequest;
+import org.eclipse.aether.resolution.ArtifactResolutionException;
+import org.eclipse.aether.resolution.ArtifactResult;
 import org.eclipse.aether.util.artifact.SubArtifact;
 import org.eclipse.aether.util.repository.AuthenticationBuilder;
 import org.guvnor.common.services.project.model.GAV;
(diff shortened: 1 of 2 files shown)


manstis:
I assume this guvnor PR is to support the ability to obtain a KJAR from the guvnor M2 Repository to then extract the image? Why not use the maven2/* URL and retrieve the KJAR as a regular HTTP GET?

mrietveld:
@manstis Re: HTTP GET -- on one hand, I like the idea, on the other hand, I think that a HTTP GET operation will be slower than a this operation. 

The other reason is that it's not really kosher (or halal? ;) ) to base a synchronous REST operation on logic that involves doing another REST operation. And I'd like to make this a synchronous REST operation. 

My reasoning is as follows -- let me know if I've missed something: 

- The current operation has the following bottleneck/performance points: 
    - It has to go through the maven repository interface
    - it does file I/O

- a HTTP GET operation has the following bottleneck/performance points:
    - it goes through HTTP (which is just as bad as file I/O, IIRC)
    - _and_ it then has the same bottlenecks as the above operation
        - it has to go through the maven repo interface
        - it does file I/O

Does that make sense?

mswiderski:
Not directly related to this change though it might be relevant - we do not store svg in kjar, only bpmn2 xml which is then used by designer

Just a side note to this PR :)

Dnia 23 maj 2015 o godz. 18:52 Marco Rietveld <notifications@github.com> napisał(a):

manstis:
Hehe.. glad I made you think and explain your design decision! The SOA
evangelists may counter argue your kosha statement.. but I won't ☺ I'll
merge on Tuesday (when I'm back to work).

> Sent when I shouldn't really be working.

@manstis <https://github.com/manstis> Re: HTTP GET -- on one hand, I like
the idea, on the other hand, I think that a HTTP GET operation will be
slower than a this operation.

The other reason is that it's not really kosher (or halal? ;) ) to base a
synchronous REST operation on logic that involves doing another REST
operation. And I'd like to make this a synchronous REST operation.

My reasoning is as follows -- let me know if I've missed something:

   -

   The current operation has the following bottleneck/performance points:
    - It has to go through the maven repository interface
      - it does file I/O
    -

   a HTTP GET operation has the following bottleneck/performance points:
    - it goes through HTTP (which is just as bad as file I/O, IIRC)
      - *and* it then has the same bottlenecks as the above operation
         - it has to go through the maven repo interface
         - it does file I/O

Does that make sense?

—
Reply to this email directly or view it on GitHub
<https://github.com/droolsjbpm/guvnor/pull/130#issuecomment-104923952>.

mrietveld:
@mswiderski we don't store the svg in the kjar.. _yet_. We will -- this is related to the designer `storesvgonsave` property.

mswiderski:
ah ok, that explains everything - thanks and sorry for side tracking :)

manstis:
Sorry Marco, I cannot merge this PR.

Adding method "File getArtifactFile( String groupId, String artifactId, String version )" to service M2RepoService breaks the ```@Remote``` service... All return types and parameters must be emulated by GWT for use on the client i.e. browser (since this is a ```@Remote``` service). java.io.File is not emulated and hence cannot be part of the ```@Remote``` interface. See http://www.gwtproject.org/doc/latest/RefJreEmulation.html#Package_java_io for details.

manstis:
If you compile all of the guvnor module (as I'm sure you did!?!?) you'd see guvnor-webapp fails for this reason.

mrietveld:
@manstis Thanks, my apologies -- I should have compiled webapp as well. I'll fix this and force update the pull.

mrietveld:
I've removed the M2RepoService* changes and only left the `GuvnorM2Repository` changes: in my code, I'll just inject the same instance of the `GuvnorM2Repository` in my own code and use the new method I created. 

I've also made sure to build all of the Guvnor repo (including webapp) this time. The build succeeded. My apologies again for not doing that previously.

manstis:
http://github.com/droolsjbpm/guvnor/commit/249cde3be

PR #131, opened 2015-06-02
Zanata changes for 6.3.0.Beta1

mbiarnes:


manstis:
http://github.com/droolsjbpm/guvnor/commit/9d052b019

psiroky:
Do we need those `# translation auto-copied from project guvnor, version 6.2.0` comments? They make the files much less readable and also the files are twice as large.

I know we should not change these property files directly as they will probably be overwritten when doing next zanata:pull, but maybe we could configure this directly in the Zanata projects?

manstis:
@psiroky Hi Petr, as you probably know we don't author the "translated" files; only the "source" (normally the us_en) ones. The content of the "translated" files is entirely down to our use of Zanata to host our i18n requirements. If you know of a configuration that can prevent the comments being inserted - but still copies the existing translations from other (Zanata) project versions then I'm all ears!

psiroky:
I am afraid my Zanata knowledge is basically zero. I might take a look into this once I find some time...I just saw something a bit unusual (the repeated comments) worth commenting, so I did comment :) It may very well be possible that this can not be easily solved.

PR #132, opened 2015-06-04
BPMSPL-141: [GSS-RFE] Ability to upload a pom-packaged artifact to Business Central

manstis:
See https://issues.jboss.org/browse/BPMSPL-141

(and a bit of a tidy-up of unused methods and constants, left over from presumably the original port from Guvnor 5.6.x).


guvnor-m2repo-editor/guvnor-m2repo-editor-api/src/main/java/org/guvnor/m2repo/service/M2RepoService.java
```diff
@@ -25,8 +25,6 @@
 @Remote
 public interface M2RepoService {
 
-    String getJarName( String path );
-
     String loadPOMStringFromJar( String path );
```
porcelli: this method doesn't support .pom files, it just throws the following when I click the "open" button of ArtifactListViewImpl:
ERROR 08-06 07:34:03,115 (GuvnorM2Repository.java:loadPOMFromJarInternal:634)     error in opening zip file

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadForm.java
```diff
@@ -107,13 +110,18 @@ public void onSubmit( final Form.SubmitEvent event ) {
                     BusyPopup.close();
                     Window.alert( M2RepoEditorConstants.INSTANCE.SelectFileUpload() );
                     event.cancel();
+                } else if ( !( isJar( fileName ) || isKJar( fileName ) || isPom( fileName ) ) ) {
```
porcelli: what about extract a method like, and just use if ( !isValid( fileName ) )
 
    private boolean isValid( final String fileName ) {
        return isJar( fileName ) || isKJar( fileName ) || isPom( fileName );
    }

manstis:
Good catch! I'll fix and update the PR.

porcelli:
http://github.com/droolsjbpm/guvnor/commit/8af533847

PR #133, opened 2015-06-16
Migrate to Lienzo 2.0.118-RC0

manstis:


manstis:
http://github.com/droolsjbpm/guvnor/commit/e20ff7be0

PR #135, opened 2015-06-29
Refactor guvnor/m2repo-editor/client/UploadForm to MVP pattern and testable form

Hasys:


guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -195,8 +146,80 @@ private String getWebContext() {
         return context;
     }
 
-    protected void showUploadingBusy() {
-        // LoadingPopup.showMessage( "Uploading...");
+    public void showUploadingBusy() {
+        // TODO: "Uploading..." needs localization
+        BusyPopup.showMessage("Uploading...");
```
Hasys: We have to add localization for this string. I can't found it in M2RepoEditorConstants class and don't know how to do it (all this localization things)

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -93,87 +95,36 @@ public WellForm doUploadForm() {
...
             }
-        } );
+        });
 
         HorizontalPanel fields = new HorizontalPanel();
-        fields.add( up );
+        fields.add(uploader);
 
         hiddenGroupIdField.setName( HTMLFileManagerFields.GROUP_ID );
-        hiddenGroupIdField.setText( null );
```
Hasys: It was duplication (about 3 times in one file) so I am moved it to separate method.

hiddenGroupIdField.setText( null );
iddenArtifactIdField.setText( null );
hiddenVersionIdField.setText( null );
hiddenFieldsPanel.setVisible( false );

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -93,87 +95,36 @@ public WellForm doUploadForm() {
...
-        //hiddenGroupIdField.setVisible(false);
```
Hasys: Some commented code removed.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -195,8 +146,80 @@ private String getWebContext() {
         return context;
     }
 
-    protected void showUploadingBusy() {
-        // LoadingPopup.showMessage( "Uploading...");
```
Hasys: This function is empty. But there was about 5-times BusyPopup.hideMessage().

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -93,87 +95,36 @@ public WellForm doUploadForm() {
...
+                presenter.handleSubmit(event);
             }
-        } );
-        up.setName( HTMLFileManagerFields.UPLOAD_FIELD_NAME_ATTACH );
+        });
 
-        form.addSubmitHandler( new Form.SubmitHandler() {
+        uploader = new FileUpload(new Command() {
             @Override
-            public void onSubmit( final Form.SubmitEvent event ) {
```
Hasys: EventHandlers was moved to presenter.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -93,87 +95,36 @@ public WellForm doUploadForm() {
...
-                    BusyPopup.close();
-                    Window.alert( M2RepoEditorConstants.INSTANCE.SelectFileUpload() );
-                    event.cancel();
-                } else if ( !( isValid( fileName ) ) ) {
-                    BusyPopup.close();
-                    Window.alert( M2RepoEditorConstants.INSTANCE.UnsupportedFileType() );
-                    event.cancel();
-                }
+            public void execute() {
+                form.submit();
```
Hasys: The only one EventHandler which wasn't moved to Presenter because the only action of this EventHandler is resend event to parent form.

guvnor-webapp/src/main/java/org/guvnor/client/perspectives/M2RepoPerspective.java
```diff
@@ -26,7 +26,8 @@
 import com.google.gwt.user.client.ui.PopupPanel;
```
Hasys: What do you think about  this file is it correct changes?

P.S. There is another 5 files with using old UploadForm:

kie-wb-distributions/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/perspectives/M2RepoPerspective.java
kie-wb-distributions/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/client/perspectives/M2RepoPerspective.java
optaplanner-wb/optaplanner-wb-webapp/src/main/java/org/optaplanner/workbench/client/perspectives/M2RepoPerspective.java
experiments/guvnor/guvnor-webapp/src/main/java/org/guvnor/client/perspectives/M2RepoPerspective.java
drools-wb/drools-wb-webapp/src/main/java/org/drools/workbench/client/perspectives/M2RepoPerspective.java

So before merge we have to change it also.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/test/java/org/guvnor/m2repo/client/upload/UploadFormTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Mock
+    private UploadFormView view;
+
+    private UploadFormPresenter uploadFormPresenter;
+
+    @Captor
+    private ArgumentCaptor<SubmitHandler> addSubmitHandler;
+    @Captor
+    private ArgumentCaptor<SubmitCompleteHandler> addSubmitCompleteHandler;
```
yurloc: @Hasys unused fields

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -62,23 +59,28 @@
...
+    private FileUpload uploader;
+
     @PostConstruct
     public void init() {
         popup.add( new ModalFooter( new Button( M2RepoEditorConstants.INSTANCE.Cancel() ) {{
             addClickHandler( new ClickHandler() {
                 @Override
                 public void onClick( ClickEvent event ) {
-                    hide();
+                    hideDialog();
```
manstis: Personally, since we preserve use of the View's super-class ```show()``` I think it's more consistent to override ```hide()``` and ```hide(boolean autoClosed)``` than introduce a ```hideDialog()``` method. Also, but you wasn't to know this, it'd be best if we extend ```BaseModel``` than ```PopupPanel``` (as the former provides consistent behaviour of all of Business Central's "popups").

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -195,8 +146,80 @@ private String getWebContext() {
...
+    }
+
+    @Override
+    public String getFileName() {
+        return uploader.getFilename();
+    }
+
+    @Override
+    public void fireSearchEvent() {
+        searchEvent.fire(new M2RepoSearchEvent());
```
manstis: Personally I try to keep event firing and observing (i.e. ```@Observes```) to the Presenter instead of the View.

guvnor-webapp/src/main/java/org/guvnor/client/perspectives/M2RepoPerspective.java
manstis: Yes this is true.. they will all need changing too. Are you to submit additional PRs?

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
manstis: Have you verified this works OK and does not affect the UI at runtime (or do you rely on the tests passing meaning everything is good?)... I just ask as I have a nagging memory that showing the ```BusyPopup``` affected the operation of the upload (e.g. it caused it not to work)... I could be wrong.. I just have this nagging recollection.

manstis: @Hasys regarding the i18n; you'll need to add a new entry to [here](https://github.com/droolsjbpm/guvnor/blob/master/guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/resources/i18n/M2RepoEditorConstants.java) and [here](https://github.com/droolsjbpm/guvnor/blob/master/guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/resources/org/guvnor/m2repo/client/resources/i18n/M2RepoEditorConstants.properties). Translation of the "default" locale happens when we push to Zanata and pull translations back into the source code later.

manstis:
So what's the plan? @Hasys Are you to adjust your commit based on @yurloc and my comments; and create PRs for the other uses of ```UploadForm``` before we merge?

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
Hasys: @manstis Yes, I am tested it at runtime as well as run new tests  and it's works as expected.
Ok, thank you I will make this changes.

guvnor-webapp/src/main/java/org/guvnor/client/perspectives/M2RepoPerspective.java
Hasys: @manstis Not yet. I will make this changes in others files if this one rewrited correctly. Just want to know is it correct changes and it's full list of files needs changes.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/test/java/org/guvnor/m2repo/client/upload/UploadFormTest.java
Hasys: @yurloc Thank you will fix it.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
yurloc: @manstis +1. For that the presenter needs to be turned into a bean (which it currently is not) and then, in M2 perspective, obtained via iocManager so that injection is performed. Do I understand that correctly?

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
Hasys: @manstis Ok, thank you. Will fix it.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
Hasys: @manstis Thank you I will rewrite it. [I didn't change it just move.]

Hasys:
@manstis yes, I will rewrite it and will send new PRs for M2Repo files as well.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
manstis: @yurloc, @Hasys .. @yurloc is correct.. At the moment I don't think this even runs (as we don't ```@Inject``` the View into the Presenter - other than manually in the test). We need to change the Presenter to ```@Inject``` the View and Event in the constructor and then instantiate the Presenter in the Perspectives via CDI... (in essence this means ```@Inject``` the UploadFormPresenter into the perspective - which will be fine as they look to be CDI-enabled already).

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -0,0 +1,202 @@
...
+                }
+            });
+        }}));
+        this.setTitle(M2RepoEditorConstants.INSTANCE.ArtifactUpload());
+        this.add(doUploadForm());
+    }
+
+    @Override
+    public void hide() {
+        uploader.getElement().setPropertyString("value", "");
```
Hasys: It's doesn't work see comment below.

guvnor-webapp/src/main/java/org/guvnor/client/perspectives/M2RepoPerspective.java
```diff
@@ -61,23 +58,16 @@ private void init() {
...
-
-                            @Override
-                            public void onClose( CloseEvent<PopupPanel> event ) {
-                                iocManager.destroyBean( uploadForm );
-                            }
-
-                        } );
-                        uploadForm.show();
+                        uploadFormPresenter.showView();
                     }
```
Hasys: Is it correct realization? Can I do not destroy it? If yes I can't to clear Filename field because there is no method in FileUpload form. So it can to be always with previous value or we have to add setFilename method to org.uberfire.ext.widgets.common.client.common.FileUpload or at least have to destroy this uploadFormView and recreate it again. What is the easiest way to solve it?

manstis: Do it one of the two ways ;)

1) Either don't ```@Inject``` it, but use ```iocManager``` to create an instance "on the fly" and then destroy it in the ```onClose``` (as per the removed code), or

2) Use ```@Inject``` and have a ```reset()``` method (or something on the Presenter) to erm, reset it -- or even have the ```showView()``` reset the form - assuming if you're showing the form you want it "reset" anyway.

guvnor-webapp/src/main/java/org/guvnor/client/perspectives/M2RepoPerspective.java
```diff
@@ -67,17 +67,18 @@ public Menus getMenus() {
                 .respondsWith( new Command() {
                     @Override
                     public void execute() {
-                        final UploadForm uploadForm = iocManager.lookupBean( UploadForm.class ).getInstance();
-                        //When pop-up is closed destroy bean to avoid memory leak
-                        uploadForm.addCloseHandler( new CloseHandler<PopupPanel>() {
+                        final UploadFormPresenter presenter = iocManager.lookupBean( UploadFormPresenter.class ).getInstance();
+                        final UploadFormViewImpl view = iocManager.lookupBean( UploadFormViewImpl.class ).getInstance();
```
manstis: New instances of the View will be created "on demand". Therefore the View instance ```@Inject```'ed into the Presenter will be a different instance to that you create here. You should really add a method to the Presenter to add a ```HideHandler``` to the View and *only* instantiate and destroy the Presenter here.

guvnor-webapp/src/main/java/org/guvnor/client/perspectives/M2RepoPerspective.java
```diff
@@ -20,13 +20,13 @@
 import javax.enterprise.event.Event;
```
Hasys: @manstis Can you verify this file? If it's correct I will send PR with another M2RepoPerspectives

manstis: @Hasys I had commented:

"New instances of the View will be created "on demand". Therefore the View instance @Inject'ed into the Presenter will be a different instance to that you create here. You should really add a method to the Presenter to add a HideHandler to the View and only instantiate and destroy the Presenter here."

But you still seem to be creating the View to add the ```HideHandler```

manstis: At the very least you should expose ```Presenter.getView()``` and then on the View interface have an ```addHideHandler```.. either way you don't want to create a new View instance to attach the ```HideHandler```.

Hasys: @manstis updated

manstis: Heh.. now you don't destroy the bean you created when the "Upload" popup closes!

You need to do something like this:

```
UploadFormPresenter uploadFormPresenter = iocManager.lookupBean( UploadFormPresenter.class ).getInstance();
uploadFormPresenter.getView().addCloseHandler( new CloseHandler<PopupPanel>() {

                            @Override
                            public void onClose( CloseEvent<PopupPanel> event ) {
                                iocManager.destroyBean( uploadFormPresenter );
                            }

                        } );
```
Or something like that!

Hasys: @manstis Sorry, forgot to say I moved it to UploadFormPresenter constructor.

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormPresenter.java
```diff
@@ -0,0 +1,109 @@
...
+    private SyncBeanManager iocManager;
+
+    @Inject
+    public UploadFormPresenter(final UploadFormView view, Event<M2RepoSearchEvent> searchEvent) {
+        this.view = view;
+        //When pop-up is closed destroy bean to avoid memory leak
+        view.addHideHandler(new HideHandler() {
+            @Override
+            public void onHide(HideEvent hideEvent) {
+                iocManager.destroyBean(view);
```
manstis: You won't need to destroy the View as this was ```@Inject```'ed by CDI, so CDI will destroy it automatically.

I worry about destroying the Presenter from "within" the Presenter too... have you tested this? The popup will call code to destroy the bean that has the code doing the destruction...

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -0,0 +1,201 @@
...
+        this.add(doUploadForm());
+    }
+
+    @Override
+    public void hide() {
+        hideGAVInputs();
+        super.hide();
+    }
+
+    public WellForm doUploadForm() {
```
manstis: Does this really need to be ```public```?

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -0,0 +1,201 @@
...
+        showWarning(M2RepoEditorConstants.INSTANCE.UnsupportedFileType());
+    }
+
+    @Override
+    public void showUploadedSuccessfullyWarning() {
+        showWarning(M2RepoEditorConstants.INSTANCE.UploadedSuccessfully());
+    }
+
+    @Override
+    public void showInvalidJarNotPomWarning() {
```
manstis: Should probably be ```showInvalidJarNoPomWarning()``` i.e. change "no*t*..." to "no..."

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormViewImpl.java
```diff
@@ -0,0 +1,201 @@
...
+        showWarning(M2RepoEditorConstants.INSTANCE.SelectFileUpload());
+    }
+
+    @Override
+    public void showUnsupportedFileTypeWarning() {
+        showWarning(M2RepoEditorConstants.INSTANCE.UnsupportedFileType());
+    }
+
+    @Override
+    public void showUploadedSuccessfullyWarning() {
```
manstis: Being picky, but is "successful upload" really a "warning"?

Would it be better to change the ```showWarning(String message)``` to be ```showMessage(...)```?

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormPresenter.java
```diff
@@ -0,0 +1,109 @@
...
+import org.jboss.errai.ioc.client.container.SyncBeanManager;
+
+@Dependent
+public class UploadFormPresenter implements UploadFormView.Presenter {
+
+    Event<M2RepoSearchEvent> searchEvent;
+
+    private UploadFormView view;
+
+    @Inject
```
manstis: Can you please move this to the constructor too (#1 for consistency, #2 you'll be able to write a test for "closing" the upload form and check ```SyncBeanManager.destroyBean()``` is called)?

Hasys:
Additional commits:
https://github.com/droolsjbpm/kie-wb-distributions/pull/75
https://github.com/droolsjbpm/drools-wb/pull/44
https://github.com/droolsjbpm/optaplanner-wb/pull/4

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormView.java
```diff
@@ -0,0 +1,59 @@
...
+        void handleSubmit(Form.SubmitEvent event);
+    }
+
+    String getFileName();
+
+    void showSelectFileUploadWarning();
+
+    void showUnsupportedFileTypeWarning();
+
+    void showUploadedSuccessfullyWarning();
```
yurloc: @Hasys pls change this to something like showUploadedSuccessfullyInfo()

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/upload/UploadFormView.java
```diff
@@ -0,0 +1,59 @@
...
+
+    void showInvalidJarNoPomWarning();
+
+    void showInvalidPomWarning();
+
+    void showUploadFailedError(String message);
+
+    void showErrorMessage(String message);
```
yurloc: This method can be private for the view impl and removed from the view interface.

manstis:
http://github.com/droolsjbpm/guvnor/commit/899f6d029

PR #136, opened 2015-07-01
Fix guvnor-m2repo-editor-backend-test-with-pom.jar for local using

Hasys:


manstis:
http://github.com/droolsjbpm/guvnor/commit/8002ae7d6

PR #137, opened 2015-07-01
GAVNOR-2206: M2Repo perspective don't filter pom.xml files.

Hasys:


manstis:
http://github.com/droolsjbpm/guvnor/commit/2e26984a1

PR #138, opened 2015-07-07
guvnor-webapp: don't build GWT permutations for IE8 and older

psiroky:
It seems we are not supporting IE8 and older (at least not in kie-wb and kie-drools-wb) so there is no need to build these permutations.


manstis:
http://github.com/droolsjbpm/guvnor/commit/f3359b485

PR #139, opened 2015-07-10
Kie builder optimizations

wmedvede:
Some changes related to optimizations provided for the kie-wb build system.
This PR basically includes some basic changes at the ResourceChangeObserver class. When we have N changes in a given project in a batch changes event, we check if the pom.xml file was changed. If it was changed then use the path to the pom.xml in the notification. Previous implementation just used the first encountered path.

### DIFF
--- a/guvnor-project/guvnor-project-builder/src/main/java/org/guvnor/common/services/builder/ResourceChangeObserver.java
+++ b/guvnor-project/guvnor-project-builder/src/main/java/org/guvnor/common/services/builder/ResourceChangeObserver.java
@@ -64,6 +64,9 @@ public class ResourceChangeObserver {
     @Any
     private Instance<ResourceChangeObservableFile> observableFiles;
 
+    @Inject
+    private ObservablePOMFile observablePomFile;
+
     public void processResourceAdd( @Observes final ResourceAddedEvent resourceAddedEvent ) {
         processResourceChange( resourceAddedEvent.getSessionInfo(),
                                resourceAddedEvent.getPath(),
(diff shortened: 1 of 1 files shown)


guvnor-project/guvnor-project-builder/src/main/java/org/guvnor/common/services/builder/ResourceChangeObserver.java
```diff
@@ -147,12 +130,40 @@ private void processResourceChange( final SessionInfo sessionInfo,
...
+
+        for ( final Map.Entry<Path, Collection<ResourceChange>> pathCollectionEntry : resourceChanges.entrySet() ) {
+
+            //Only process Project resources
+            project = projectService.resolveProject( pathCollectionEntry.getKey() );
+            if ( project == null ) {
+                continue;
+            }
+
+            if ( !pendingNotifications.containsKey( project ) && isObservableResource( pathCollectionEntry.getKey() ) ) {
```
manstis: Won't POM files be detected here; and the ```else``` block be skipped?

wmedvede: Not necessary, a bach changes event can contain changes for multiple files within the same project.
What we want is: "if the pom.xml file is in the set of changed files , then always keep the path to the pom.xml and put this path in the event", if not, put whatever any other path.

wmedvede: if first file for a given project is TheFile.java, the this file is putted on the hash, but then if pom.xml is processed then we want to remember it instead of the initial TheFile.java.

wmedvede: This will be useful for the optimizations at kie-wb-common build system.

manstis:
http://github.com/droolsjbpm/guvnor/commit/5aae6fe14

PR #140, opened 2015-07-15
Update zeroclipboard src

cristianonicolai:


manstis:
I think their removal may be fixing the symptoms of a problem.

These JS files are needed by the guvnor-structure-client module; for the Repositories List View, see here: https://github.com/droolsjbpm/guvnor/blob/master/guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/list/RepositoriesViewItem.java#L160 and the JS files are included here https://github.com/droolsjbpm/guvnor/tree/master/guvnor-structure/guvnor-structure-client/src/main/resources/org/guvnor/structure/public/zeroclipboard.

If they are removed I suspect the "Copy URL" feature on the Repository List View will stop working.

The JS could not be injected with a &lt;script&gt; tag in the guvnor-stucture-client's gwt.xml file as they need to be in the &lt;HEAD&gt; for ZeroClipboard to bootstrap correctly.

manstis:
BTW, I only mention on guvnor as you'll find the same JS is injected into the &lt;HEAD&gt; in drools-wb, optaplanner-wb, kie-drools-wb and kie-wb. The "Copy URL" feature was broken for a long time (as ```$wnd.ZeroClipboard``` was undefined) and I went through the webapps adding the JS injection.

manstis:
OK, the *real* problem is that the JS lives in ```/org.guvnor.GuvnorWorkbench/zeroclipboard/``` and not ```/org.guvnor.GuvnorWorkbench/js/```

manstis:
@cristianonicolai Thanks.. I typed the above just as you'd updated the PR.. merging now.

cristianonicolai:
Thanks for the info @manstis :) Ok, so the problem with zeroclipboard was actually a wrong path. I fixed it in this commit.

manstis:
http://github.com/droolsjbpm/guvnor/commit/19870232b

manstis:
BTW, @psiroky and me were having a discussion about ```guvnor-webapp``` with @mdproctor.. Whilst we know *what* it's meant to be (a showcase for generic Project features: building, deploying, Asset Management, OU Management, Repository Management etc) guvnor-webapp is really (a) feature incomplete and (b) not maintained a great deal.. LiveSpark is the current "preferred demo webapp" of choice (other than UFs!!) and (ideally) guvnor-webapp and guvnor-webapp-distribution-wars could be removed (for now).

PR #142, opened 2015-07-15
clone repo form refacotred to MVP

jomarko:
PR for jira https://issues.jboss.org/browse/GUVNOR-2218


guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryFormTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @Mock
+    private OrganizationalUnit ouUnit1;
+    @Mock
+    private OrganizationalUnit ouUnit2;
+    @Mock
+    private Repository repository;
+    @Captor
+    private ArgumentCaptor<Boolean> boolArgument;
+    @InjectMocks
+    private CloneRepositoryPresenter presenter = new MockCloneRepositoryPresenter(false);
```
Rikkola: Nice, I just learned some thing new. I have been using the constructor for setting up the injected fields. This feels a lot nicer.

guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/mock/MockCloneRepositoryPresenter.java
```diff
@@ -0,0 +1,21 @@
+package org.guvnor.structure.client.editors.repository.clone.mock;
+
+import org.guvnor.structure.client.editors.repository.clone.CloneRepositoryPresenter;
+
+/**
+ * Created by jomarko on 7/15/15.
+ */
```
Rikkola: We try to keep the code free of author tags for few reason that are explained here. 
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/README.md#faq
But I don't think the PR should be rejected because of this. This just gets removed once somebody spots it.

Rikkola:
Over all nice looking MVP code. Good to see the code base getting better.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+        return true;
+    }
+
+    protected void setRepositoryService(Caller<RepositoryService> repositoryService) {
+        this.repositoryService = repositoryService;
+    }
+
+    protected void setOrganizationalUnitService(Caller<OrganizationalUnitService> organizationalUnitService) {
+        this.organizationalUnitService = organizationalUnitService;
+    }
```
Rikkola: I'm guessing you have these two setter methods because of unit testing. The cleaner way is to just do the injections with the constructor. Here is an example: 
https://github.com/droolsjbpm/drools-wb/blob/a88de6106110ae000b72e6dd44a40b7ab746d869/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/KSessionSelector.java#L52-52

jomarko: Yes, it is because of unit testing, I will use constructor injection by your proposal. I will also remove author annotations. Then I will amend my commit. Ok?

jomarko:
removed author comments

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
+package org.guvnor.structure.client.editors.repository.clone;
```
manstis: Needs license header.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryView.java
```diff
@@ -0,0 +1,87 @@
+package org.guvnor.structure.client.editors.repository.clone;
```
manstis: Needs license header.

guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryFormTest.java
```diff
@@ -0,0 +1,121 @@
+package org.guvnor.structure.client.editors.repository.clone;
```
manstis: Needs license header.

guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/answer/OuServiceAnswer.java
```diff
@@ -0,0 +1,41 @@
+package org.guvnor.structure.client.editors.repository.clone.answer;
```
manstis: Needs license header.

guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/answer/RepositoryServiceAnswer.java
```diff
@@ -0,0 +1,39 @@
+package org.guvnor.structure.client.editors.repository.clone.answer;
```
manstis: Needs license header.

guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/mock/MockCloneRepositoryPresenter.java
```diff
@@ -0,0 +1,18 @@
+package org.guvnor.structure.client.editors.repository.clone.mock;
```
manstis: Needs license header.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+        view.setPasswordEnabled(true);
+        view.setUsernameEnabled(true);
+        view.setGitUrlEnabled(true);
+        view.setOrganizationalUnitEnabled( true );
+        view.setNameEnabled(true);
+    }
+
+    protected boolean isOuMandatory() {
+        try {
+            final IOCBeanDef<RepositoryPreferences> beanDef = IOC.getBeanManager().lookupBean(RepositoryPreferences.class);
```
manstis: You can ```@Inject``` an instance of ```RepositoryPreferences``` rather than use ```IOC``` here. The old code used ```IOC``` as the old code was not CDI-enabled and hence we needed to instantiate the ```RepositoryPreferences``` object with the IOC Bean Manager manually.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+    @Override
+    public void handleCancelClick() {
+        view.hideForm();
+    }
+
+    @Override
+    public void handleCloneClick() {
+        if ( view.isGitUrlEmpty() ) {
+            view.setUrlGroupType( ControlGroupType.ERROR );
+            view.setUrlHelpMandatory();
```
manstis: ```view.showUrlHelpManatoryMessage()``` would be better. You're not *setting* anything.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+            return;
+        } else if ( !URIUtil.isValid(view.getGitUrl()) ) {
+            view.setUrlGroupType( ControlGroupType.ERROR );
+            view.setUrlHelpInvalidFormat();
```
manstis: ```view.showUrlHelpInvalidFormatMessage()``` would be better. You're not *setting* anything.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+            return;
+        } else {
+            view.setUrlGroupType( ControlGroupType.NONE );
+        }
+        final String organizationalUnit = view.getOrganizationalUnit(view.getSelectedOrganizationalUnit());
+        if ( isOuMandatory && !availableOrganizationalUnits.containsKey( organizationalUnit ) ) {
+            view.setOrganizationalUnitGroupType( ControlGroupType.ERROR );
+            view.setOrganizationalUnitHelpMandatory();
```
manstis: ```view.showOrganizationalUnitHelpMandatoryMessage()``` would be better. You're not *setting* anything.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+            return;
+        } else {
+            view.setOrganizationalUnitGroupType( ControlGroupType.NONE );
+        }
+
+        if ( view.isNameEmpty() ) {
+            view.setNameGroupType( ControlGroupType.ERROR );
+            view.setNameHelpMandatory();
```
manstis: ```view.showNameHelpManatoryMessage()``` would be better. You're not *setting* anything.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+                    final String password = view.getPassword().trim();
+                    final Map<String, Object> env = new HashMap<String, Object>( 3 );
+                    env.put( "username", username );
+                    env.put( "crypt:password", password );
+                    env.put( "origin", origin );
+
+                    repositoryService.call( new RemoteCallback<Repository>() {
+                                                @Override
+                                                public void callback( Repository o ) {
+                                                    view.closeBusyPopup();
```
manstis: We have an interesting scenario here. If you show the form once and successfully create a Repository the form remains "locked" due to the call to ```lockScreen()``` above. The screen is successfully "unlocked" in the ```ErrorCallback``` but not in the ```RemoteCallback```. You do not "unlock" the screen in the ```showForm()``` method so I think there's a risk the popup could remain locked.

manstis: You never call ```view.showBusyPopupMessage()``` so why hide it?

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+    protected boolean isOuMandatory() {
+        try {
+            final IOCBeanDef<RepositoryPreferences> beanDef = IOC.getBeanManager().lookupBean(RepositoryPreferences.class);
+            return beanDef == null || beanDef.getInstance().isOUMandatory();
+        } catch ( IOCResolutionException exception ) {
+        }
+        return true;
+    }
+
+    protected void setRepositoryService(Caller<RepositoryService> repositoryService) {
```
manstis: Agree with @Rikkola ... try to have these on the constructor.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
```diff
@@ -0,0 +1,211 @@
...
+        this.repositoryService = repositoryService;
+    }
+
+    protected void setOrganizationalUnitService(Caller<OrganizationalUnitService> organizationalUnitService) {
```
manstis: Agree with @Rikkola ... try to have these on the constructor.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryView.java
manstis: Whilst we're trying to improve out code quality (and test coverage) JavaDocs are always nice to add *now* rather than *later*.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryViewImpl.java
```diff
@@ -0,0 +1,322 @@
...
+import com.google.gwt.user.client.ui.Widget;
+import org.guvnor.structure.client.resources.i18n.CommonConstants;
+import org.uberfire.ext.widgets.common.client.common.BusyPopup;
+import org.uberfire.ext.widgets.common.client.common.popups.BaseModal;
+import org.uberfire.ext.widgets.common.client.common.popups.errors.ErrorPopup;
+import org.uberfire.ext.widgets.core.client.resources.i18n.CoreConstants;
+
+@Dependent
+public class CloneRepositoryViewImpl
+        extends PopupPanel implements CloneRepositoryView {
```
manstis: This should extend ```BaseModal``` and not ```PopupPanel```. ```BaseModal``` provides a standard look and feel to the popups we use in the Workbench.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryViewImpl.java
```diff
@@ -0,0 +1,322 @@
...
+    HelpInline urlHelpInline;
+
+    @UiField
+    TextBox usernameTextBox;
+
+    @UiField
+    PasswordTextBox passwordTextBox;
+
+    @UiField
+    BaseModal popup;
```
manstis: OK, this is a classic wrapping a "popup" in a "popup" mistake.

Your custom widget extends ```PopupPanel``` (which is a popup) and you embed another ```BaseModal``` (another popup implementation) in the Widget's composite content. You then show/hide the embedded ```BaseModal```. You should be able to simplify this by having your Widget extend ```BaseModal``` and remove the wrapped popup.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryView.java
```diff
@@ -0,0 +1,87 @@
...
+
+    void setPasswordEnabled(boolean enabled);
+
+    void setCloneEnabled(boolean enabled);
+
+    void setCancelEnabled(boolean enabled);
+
+    void setPopupCloseVisible(boolean closeVisible);
+
+    void showBusyPopupMessage();
```
manstis: You don't call this anywhere.. perhaps you should, before the call to the server to clone a Repository?

The implementation of this *normally* uses ```BusyPopup``` which defers is display until after 250ms.. so you can "show" it safe in the knowing if the server-side operation completes in <250ms the ```BusyPopup``` will not be shown to Users (provided you obviously ```closeBusyPopup```!)

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryView.java
```diff
@@ -0,0 +1,87 @@
+package org.guvnor.structure.client.editors.repository.clone;
+
+import com.github.gwtbootstrap.client.ui.constants.ControlGroupType;
+
+public interface CloneRepositoryView {
```
manstis: You can extend ```org.uberfire.ext.widgets.common.client.common.HasBusyIndicator``` (which has the methods to show/hide a "<doing something>..." popup message to Users (instead of having ```showBusyPopupMessage()``` and ```closeBusyPopup()```).

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryViewImpl.java
```diff
@@ -0,0 +1,322 @@
...
+    public void init(Presenter presenter, boolean isOuMandatory) {
+        this.presenter = presenter;
+
+        setWidget(uiBinder.createAndBindUi(this));
+
+        if ( !isOuMandatory ) {
+            isOUMandatory.removeFromParent();
+        }
+
+        popup.setDynamicSafe( true );
```
manstis: You can remove this if you extend ```BaseModal```.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryViewImpl.java
```diff
@@ -0,0 +1,322 @@
...
+            @Override
+            public void onKeyPress(final KeyPressEvent event) {
+                urlGroup.setType(ControlGroupType.NONE);
+                urlHelpInline.setText("");
+            }
+        });
+    }
+
+    @Override
+    public void hideForm() {
```
manstis: If you extend ```BaseModal``` you can override the ```show()``` method instead.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryViewImpl.java
```diff
@@ -0,0 +1,322 @@
...
+        BusyPopup.close();
+        popup.hide();
+        super.hide();
+    }
+
+    @Override
+    public void showForm() {
```
manstis: No need to do this if you extend ```BaseModal``` (just change the View interface method name to ```show()```).

guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/mock/MockCloneRepositoryPresenter.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.guvnor.structure.client.editors.repository.clone.CloneRepositoryPresenter;
+
+public class MockCloneRepositoryPresenter extends CloneRepositoryPresenter {
+
+    private boolean isOuMandatory;
+
+    public MockCloneRepositoryPresenter(boolean isOuMandatory) {
```
manstis: If you change to use constructor injection in the Presenter you will not need to provide a "mock" method overriding the ```isOUMandatory()``` method as you can pass a mock ```RepositoryPreferences``` object to the constructor and set it up to return true/false depending on the test.

jomarko:
added licences

jomarko:
renamed set methods to show methods

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryView.java
jomarko: Not applicable because of CoreConstants used in view. HasBusyIndicator interface has method for showing busy state, which takes message parameter. This parameter is useless because I use CoreConstants in view.

jomarko:
used BaseModal as base class for view implementation

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
Rikkola: Sure. Doesn't have to be constructor injection if you figure out another way. The goal is just to keep code that is used for testing on the test side and out of the main side.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryView.java
manstis: OK. Understood.

manstis:
Great, I think we are getting closer to be able to merge.

I still would prefer to ```@Inject``` an instance of ```RepositoryPreferences``` rather than use the ```IOC``` class in ```CloneRepositoryPresenter```. See my original comment in the PR.

Once this has been changed we will need additional PRs for, drools-wb, livespark, optaplanner-wb and kie-wb-distributions repositories to fix the instantiation of ```CloneRepositoryForm``` (now ```CloneRepositoryPresenter```) like you've done in ```GuvnorWorkbenchEntryPoint```.

In the listed repositories, look at the xxx-webapp module and the ```AdministrationPerspective``` for example: https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-webapp/src/main/java/org/drools/workbench/client/perspectives/AdministrationPerspective.java#L121

jomarko:
@manstis please, stay tuned, I am not at work to Monday.

jomarko:
done

jomarko:
added addCloseHandler method to presenter

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryViewImpl.java
```diff
@@ -0,0 +1,309 @@
...
+import org.uberfire.ext.widgets.common.client.common.popups.errors.ErrorPopup;
+import org.uberfire.ext.widgets.core.client.resources.i18n.CoreConstants;
+
+@Dependent
+public class CloneRepositoryViewImpl
+        extends BaseModal implements CloneRepositoryView {
+
+    @Override
+    public HandlerRegistration addCloseHandler(CloseHandler handler) {
+        return null;
```
manstis: IDK why you've added support for ```CloseHandler``s but this is going to cause problems for sure.

I'll merge these PRs when we've fixed this new issue...

jomarko: Hmm, I forgot about it, It generated my IDE. I add it to interface because of this https://github.com/jomarko/optaplanner-wb/blob/RHBXMSQE-416/optaplanner-wb-webapp/src/main/java/org/optaplanner/workbench/client/perspectives/AdministrationPerspective.java#L123

manstis:
http://github.com/droolsjbpm/guvnor/commit/b8d352790

manstis:
Thanks for doing this.. I hope you've learned a lot from the experience :)

manstis:
http://github.com/droolsjbpm/guvnor/commit/02f6fc2f7

PR #144, opened 2015-07-19
[JBPM-4703] Can't create new project containing spaces in name via REST API

abhijithumbe:
Proposal fix for JBPM-4703 . Please review.

### DIFF
--- a/guvnor-rest/guvnor-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
+++ b/guvnor-rest/guvnor-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
@@ -207,7 +207,7 @@ public class JobRequestHelper {
             return result;
         } else {
             POM pom = new POM();
-            pom.getGav().setArtifactId( projectName );
+            pom.getGav().setArtifactId( projectName.replaceAll("\\s","") );
             pom.getGav().setGroupId( projectGroupId );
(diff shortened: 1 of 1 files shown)


manstis:
@mrietveld We use this https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/wizard/NewProjectWizard.java#L152 to sanitize ArtifactId's when creating new Projects from the UI. It'd probably be best to use this for sanitizing ArtifactsId's when creating new Projects with REST.

mrietveld:
@manstis but, in the UI code, basically the project name is *changed* to a new one without spaces, right? 

Of is this just something that's done underwater -- and if it's done underwater, what happens when both "My Proj Name" and "MyProjName" exist?

manstis:
1) Yes, it removes all characters not [A-Za-z0-9_\-.] not just spaces.

2) Yes it is underwater (but the User is given a chance to "see" the sanitized name before committing)

3) If the sanitized name already exists a VFS error is thrown (and shown to users).

porcelli:
@mrietveld what's the status of it? still relevant?

mrietveld:
Please close this: the intention behind the patch is not possible. 

For more information, see https://bugzilla.redhat.com/show_bug.cgi?id=1243693

PR #146, opened 2015-07-22
[RHBXMSQE-418] - added testcases

jomarko:


jomarko:
please do not merge PR yet, I am adding more testcases

jomarko:
@manstis , please do you have any idea, why in this test [1], in presenter URIUtil.isValid("abc") returns true, but if I run web-app and I fill "abc" as Git URL, I obtain message about invalid url format.

[1]
https://github.com/jomarko/guvnor/blob/RHBXMSQE-418/guvnor-structure/guvnor-structure-client/src/test/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryFormTest.java#L239

manstis:
@jomarko It is probably because when compiled into the GWT application we substitute the ```URIUtil``` class with this one https://github.com/uberfire/uberfire/blob/master/uberfire-api/src/main/resources/org/uberfire/jre/org/uberfire/util/URIUtil.java which delegates to a JavaScript library; whereas when run as a test it's using https://github.com/uberfire/uberfire/blob/master/uberfire-api/src/main/java/org/uberfire/util/URIUtil.java.

jomarko:
Desired testcases completed.

manstis:
I squashed the commits into one and merged.

http://github.com/droolsjbpm/guvnor/commit/a6355c362

When you next submit a PR (assuming you might look at the "Create Repository" form too; can you please be sure to sync your fork with origin/master? This PR contained changes that were different to what it at origin/master (when I fixed the CDI problem yesterday I made a few formatting changes - i.e. I used our default code-style formatting before committing).

jomarko:
Sorry that my inattention caused additional work for you. Thanks for merge. Sure I can do the same with "Create Repository Form".

triceo:
@manstis - when it comes to the ```URIUtil``` class and the fact that it is different between tests and production... is this standard practice, something that we're recommending to do? (It concerns me - because this is something that forces us to continue using Selenium.)

manstis:
@triceo It's not standard practice for sure.. ```URIUtil``` is the only case I know of.

Looking at the code it appears the class is made available to both client (browser) and server side; however the client side implementation is "overlayed" with an implementation that uses http://github.com/garycourt/uri-js instead of ```java.net.URLEncoder``` (that is not available client side).

@porcelli was the originator of this code.. if anything the issue is that there is a difference in behaviour of the two implementations.. which really should not be the case. The ideal solution is to obviously ensure both implementations are consistent.

mrietveld:
@manstis, @triceo: there's an old issue with the `URIUtil` -- it was in a library that we stopped using, I think? We created our own local copy, I think? 

That's probably part of the reason for the issue with `URIUtil` here.

PR #158, opened 2015-09-14
Remove guvnor-webapp and distribution-wars on 6.3.x

psiroky:
This removes the `guvnor-webapp` and `guvnor-webapp-distribution-wars` modules from _6.3.x_ branch. They does not seem to be used and only considerably prolong the build. They will stay on master for now as they may be needed in future.

@Rikkola, @manstis, @porcelli, @wmedvede, @csadilek, @ederign  please let me know in case you are using it and don't want it to be removed.

Let's wait with the merge till we hear from the others.


cristianonicolai:
@psiroky During the PatternFly migration, I found quite useful to have apps like this and the showcases apps. Specially for testing and using module specific features without having to build a series of modules. I understand it does ads extra code for maintenance and build time, so Im not sure about the others team member opinion but maybe one suggestion would be to remove the apps from the maven default profile, but leave it there so it could be used if needed.

psiroky:
@cristianonicolai thanks for the feedback. I think that if we decide to leave the code there, we should also build it periodically (even if only the FastCompiled version). That would make sure the webapp/showcase is at least compilable. From my experience -- if the webapp is not part of the build (so that it's compilation won't cause build failure), it is very likely to get forgotten and changes won't be done there (because the build is not failing). Having the build failures reported early will definitely decrease the maintenance cost. From my point of view, there should be no "dead" code, meaning code that is not being built. Everything that is in the repo should be built, otherwise it gets stale very fast.

psiroky:
Closing this one as the webapp/showcase is being used.

PR #161, opened 2015-09-23
BZ1263454: Unable to create Project in cloned Git Repository

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1263454

The problem was that the ```CreateRepositoryForm``` did not set the repository to be "unmanaged".

I fixed that here (and moved the constants in ```EnvironmentParameters``` to the ```-api``` module to allow re-use of the constants on both the server and client - as it had been duplicated in ```CreateRepositoryWizard```).


manstis:
*guvnor*
(6.3.x) http://github.com/droolsjbpm/guvnor/commit/a32337e68

PR #163, opened 2015-10-13
asset-mgmt - missing definition of itemDefinition causes parser problems

mswiderski:


psiroky:
+1. I tried locally and it fixes the failing kie-wb-distributions smoke tests.

PR #166, opened 2015-10-21
BZ1265901: NullPointerException with kie-config-cli.sh after migration from 6.1.0 to 6.1.2

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1265901

The cause was that if the "old" ```ConfigGroup``` did not have a ```security:roles``` entry a "new" ```security:groups``` entry was not created; however ```RepositoryServiceImpl.addGroup(...)``` always expects a non-null ```security:groups``` entry.


manstis:
@porcelli Feel free to re-assign (but this is, again, somewhere you know about).

manstis:
*guvnor*
(master) http://github.com/droolsjbpm/guvnor/commit/85f139e85

PR #167, opened 2015-10-21
BZ-1273109 - BPM cluster: Job executor fails with NPE while

mswiderski:
this PR enables efficient and reliable job status checks within cluster. Since all the commands/jobs are executed asynchronously (via processes and jbpm executor) they might be executed on any cluster member. Thus when running clustered environment default map based cache did not work properly when request for checking job status was sent to other instance then the job was actually executed.
This PR uses executorService (component that actually executed the job) to find it in db in case it does not exist in the local cache. That way job status check is cluster aware and will cover efficient job execution distribution across cluster members while still allowing to find the job status regardless where it was executed. 
Especially important when kie-wb instances are running behind load balancer.

ExecuteOperation process was enhanced to allow to pass custom business key that is then used to find the job to check it status. This business key is actually job id so it provides uniqueness and thus can be reliably searched.

this is one of three PRs for this BZ - others are in jbpm (https://github.com/droolsjbpm/jbpm/pull/317) and droolsjbpm-integration (https://github.com/droolsjbpm/droolsjbpm-integration/pull/212), though they don't have any compile dependency

### DIFF
--- a/guvnor-asset-mgmt/guvnor-asset-mgmt-project/src/main/resources/ExecuteOperation.bpmn2
+++ b/guvnor-asset-mgmt/guvnor-asset-mgmt-project/src/main/resources/ExecuteOperation.bpmn2
@@ -9,6 +9,7 @@
   <bpmn2:itemDefinition id="_JobRequestItem" structureRef="org.guvnor.rest.client.JobRequest"/>
   <bpmn2:itemDefinition id="_RetriesItem" structureRef="Integer"/>
   <bpmn2:itemDefinition id="_OwnerItem" structureRef="String"/>
+  <bpmn2:itemDefinition id="_BusinessKeyItem" structureRef="String"/>
   <bpmn2:error id="java.lang.RuntimeException" errorCode="java.lang.RuntimeException"/>
   <bpmn2:itemDefinition id="__5D7392FA-7863-4ECE-AE2E-2A55B32CA6E8_OperationInputXItem" structureRef="Object"/>
   <bpmn2:itemDefinition id="__5D7392FA-7863-4ECE-AE2E-2A55B32CA6E8_RequestorInputXItem" structureRef="Object"/>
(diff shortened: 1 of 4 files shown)


mrietveld:
+1 Feel free to merge. 

@mswiderski Could you describe integration tests cases in a comment here? Then I can implement them in the kie-test modules or in the smoke tests.

manstis:
Just asking; why do you use jobExecutor.get() rather than iterate the instances - if there's more than one surely any could know of a jobs' status?

I'm also going to pretend to be clever, but should any of the remote lookup via job executor be synchronized?

mswiderski:
@mrietveld integration test is rather difficult to prepare as it requires cluster setup. There must be at least two servers in domain mode that make this to be used. As you can see it first checks the local cache and will look up via executor (which is essentially db query) only if it not exists in local cache or it exists but has status accepted. So to create test you would need to start to eap/wildfly instances as part of the test. Is that feasible?
Anyway, test should perform any operation (like project compile or install) and then check on both server instances that status of the job. Successful result is when both server instances will reply with successful job completion response for the same job id

mswiderski:
@manstis the use of Instance<ExecutorService> was done only to make it optional when running guvnor webapp or drools-wb which most likely do no have executor service (jbpm dependencies) available. There cannot be more than on executor service in workbenches at the moment so iterating over them won't make any difference.
As it comes to synchronization, not sure I see a need for it. Executor service will perform db query so as it's only read operation it does not affect any state. The only "write" operation is to put the job found in executor into the local cache though regardless if many threads will perform this query at the same time they will always return exact same result. Or did I miss something here?

manstis:
@mswiderski Thanks for the explanation :) I see Jobs is a SynchronizedMap so the only fear I had is misplaced anyway. I'll get back to my TextBoxes, Buttons and ListBoxes...

PR #173, opened 2015-10-29
webapp: tweak and fix GWT config

psiroky:
* define the supported user agents via property. This will
   ultimately enable sharing the property for all webapps/showcase
 * remove suport for IE10 and older. Only IE11+ is supported
 * do not use 'resources' goal of gwt-maven-plugin. It does not play
   well with filtered resources and replaces the files filtered
   by standard maven-resources-plugin. The 'resources' goal should
   not be needed after additional config in pom

I've tested this by running `mvn clean install`, `mvn clean install -Dfull` and `mvn clean gwt:run` and everything seems to be working fine. 

@manstis, @porcelli, @Rikkola, @csadilek please take a look and let me know I missed something (the only thing I am not 100% sure about is the removal of 'resources' goal from the gwt-maven-plugin).


guvnor-webapp/pom.xml
```diff
@@ -191,22 +192,6 @@
...
-    </dependency>
-
-    <dependency>
-      <groupId>org.guvnor</groupId>
-      <artifactId>guvnor-asset-mgmt-client</artifactId>
-      <scope>provided</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.guvnor</groupId>
```
psiroky: These `asset-mgmt` deps were duplicated (they are declared again later in the pom).

guvnor-webapp/src/main/resources/log4j.xml
```diff
@@ -1,43 +0,0 @@
...
-    </layout>
-  </appender>
-  <logger name="org.drools">
-    <level value="INFO"/>
-  </logger>
-  <root>
-    <priority value="ERROR"/>
-    <appender-ref ref="CONSOLE"/>
-  </root>
-</log4j:configuration>
```
psiroky: We should not be using `log4j` at all. This config file is useless.

csadilek:
@psiroky skipping the resource goal in the gwt-maven-plugin is good actually. See these commits which fixed refreshes in some IDEs: https://github.com/errai/errai-tutorial/commit/729d98e648062c76746d4ff773b0a396213e657c and https://github.com/uberfire/uberfire/commit/56e230739315472258bed925f61d7787525c27b8

csadilek:
+1, but let's give others time to take a look as well.

psiroky:
Great. We definitely want to keep the execution of `maven-resources-plugin` as it will replace the `gwt.user.client.all` property with the actual value.

manstis:
@psiroky Hi, my only concern would be whether [this](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-widgets/kie-wb-common-ui/pom.xml#L228) is still handled correctly. That said it seems removal of the maven-gwt-compiler resources goal is preferential elsewhere so, if it breaks, we'd be better finding another alternative.

psiroky:
@manstis it should still work correctly as that module is not using the GWT `resources` goal. If I understand correctly, it relies on standard `maven-resources-plugin` to filter and copy the resources.

csadilek:
Thanks.

master: 1d5c8c323f6b7342e8ae8d4d3fb6b69085da53e7

psiroky:
Thanks @csadilek. I will create similar PRs for other repos as well.

PR #175, opened 2015-11-03
webapp: use shared GWT properties + tweak memory settings

psiroky:
* gwt.user.agent.all property is inherited from kie-parent-metadata
 * memory settings adjusted as there is no need
   for such high values. Tested with "mvn clean install -Dfull"

Needs to be merged together with/after https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/93

### DIFF
--- a/guvnor-webapp/pom.xml
+++ b/guvnor-webapp/pom.xml
@@ -37,7 +37,6 @@
     <as.version>8.1.0.Final</as.version>
     <!-- Add the absolute path for $JBOSS_HOME below to manage another instance -->
     <errai.jboss.home>${project.build.directory}/wildfly-${as.version}</errai.jboss.home>
-    <gwt.user.agent.all>gecko1_8,safari</gwt.user.agent.all>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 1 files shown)


manstis:
Where's the gwt.compiler.localWorkers parameter used? (Probably obvious but I could not see on my phone).

psiroky:
It is used in `gwt-maven-plugin` configuration. Currently in `kie-parent-metadata`. See https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/pom.xml#L165. The advantage is that we are not duplicating the config, but as you pointed out it may be a bit confusing as it is not used in the POM that declares the property. What we could do is to specify the config in here as well (e.g. add `<localWorkers>${gwt.compiler.localWorkers}</localWorkers>` into `gwt-maven-plugin` config. It would be duplicated, but the value would not change, so that should be fine. WDYT?

manstis:
Nope OK as it is; I just couldn't readily see where it was used.

+1 to merge.

psiroky:
I am still considering it, because you have good point -- it is not completely obvious where the property is used, which is definitely not ideal.

psiroky:
OK, let's leave it as is. We could change it later in case of wider confusion.

manstis:
+1 to merge then (in case my original +1 is lost in the narrative)

psiroky:
I've slightly updated the PR by removing (now) useless configuration of eclipse `lifecycle-mapping` plugin (it was configured to ignore the gwt-maven-plugin `resources` goal, but that has been removed altogether).

csadilek:
Thanks!

master: 9b72874310e8ce95cca1b450915125eb5b3f0693

PR #178, opened 2015-11-04
- BZ-1274691: fixing view checks

Salaboy:
You can now release only from branches starting with release*.
Modifications only on the view, I've tried GWT Mockito but there were several issues with View Elements not working with GWT mockito

### DIFF
--- a/guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/release/ReleaseScreenPopupPresenter.java
+++ b/guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/release/ReleaseScreenPopupPresenter.java
@@ -1,7 +1,17 @@
 /*
- * To change this license header, choose License Headers in Project Properties.
- * To change this template file, choose Tools | Templates
- * and open the template in the editor.
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 package org.guvnor.asset.management.client.editors.repository.structure.release;
 
--- a/guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/release/ReleaseScreenPopupPresenter.java
+++ b/guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/release/ReleaseScreenPopupPresenter.java
@@ -1,0 +1,101 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.guvnor.asset.management.client.editors.repository.structure.release;
+
+import com.github.gwtbootstrap.client.ui.constants.ControlGroupType;
+import com.google.gwt.user.client.Command;
+import org.guvnor.asset.management.client.i18n.Constants;
+
+
+public class ReleaseScreenPopupPresenter {
+     
+    private ReleaseScreenPopupViewImpl view;
+    private Command callbackCommand;
+
+    public ReleaseScreenPopupPresenter(ReleaseScreenPopupViewImpl view, Command callbackCommand) {
+        this.view = view;
+        this.callbackCommand = callbackCommand;
+    }
+    
+    
+    
+    private final Command okCommand = new Command() {
+        @Override
+        public void execute() {
+            if ( isEmpty( view.getVersionText().getText() ) ) {
+                view.getVersionTextGroup().setType( ControlGroupType.ERROR );
+                view.getVersionTextHelpInline().setText( Constants.INSTANCE.FieldMandatory0( "Version" ) );
+
+                return;
+            }
+            if ( isSnapshot( view.getVersionText().getText() ) ) {
+                view.getVersionTextGroup().setType( ControlGroupType.ERROR );
+                view.getVersionTextHelpInline().setText( Constants.INSTANCE.SnapshotNotAvailableForRelease( "-SNAPSHOT" ) );
+
+                return;
+            }
+            if( !view.getSourceBranchText().getText().startsWith("release")){
+                view.getSourceBranchTextGroup().setType( ControlGroupType.ERROR );
+                view.getSourceBranchTextHelpInline().setText( Constants.INSTANCE.ReleaseCanOnlyBeDoneFromAReleaseBranch());
+                return;
+            }
+            if ( view.getDeployToRuntimeCheck().getValue() ) {
+
+                if ( isEmpty( view.getUserNameText().getText() ) ) {
+                    view.getUserNameTextGroup().setType( ControlGroupType.ERROR );
+                    view.getUserNameTextHelpInline().setText( Constants.INSTANCE.FieldMandatory0( "Username" ) );
+
+                    return;
+                }
+
+                if ( isEmpty( view.getPasswordText().getText() ) ) {
+                    view.getPasswordTextGroup().setType( ControlGroupType.ERROR );
+                    view.getPasswordTextHelpInline().setText( Constants.INSTANCE.FieldMandatory0( "Password" ) );
+
+                    return;
+                }
+
+                if ( isEmpty( view.getServerURLText().getText() ) ) {
+                    view.getServerURLTextGroup().setType( ControlGroupType.ERROR );
+                    view.getServerURLTextHelpInline().setText( Constants.INSTANCE.FieldMandatory0( "ServerURL" ) );
+
+                    return;
+                }
+
+            }
+
+            if ( callbackCommand != null ) {
+                callbackCommand.execute();
+            }
+            view.hide();
+        }
+
+        private boolean isEmpty( String value ) {
+            return value == null || value.isEmpty() || value.trim().isEmpty();
+        }
+
+        private boolean isSnapshot( String value ) {
+            return value != null && value.trim().endsWith( "-SNAPSHOT" );
+        }
+    };
+
+    private final Command cancelCommand = new Command() {
+        @Override
+        public void execute() {
+            view.hide();
+        }
+    };
+
+    public Command getOkCommand() {
+        return okCommand;
+    }
+
+    public Command getCancelCommand() {
+        return cancelCommand;
+    }
+    
+    
+}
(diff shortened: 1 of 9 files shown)


yurloc:
Yes, adding basic tests would require some refactoring work towards MVP style. I can see some validation logic and `Command`s that should be in a presenter etc. so the view impl could be mocked out in the test.

I would vote for using this opportunity to do the refactoring and including tests in this PR.

Salaboy:
@yurloc yeah, if I move the command to the Presenter a test can be quickly written. I will try to add that asap

porcelli:
@Salaboy you don't need it. so... if MVP is correctly implemented, you don't need to test view code, as logic will be on the presenter. 
-1 until we get MVP + test

manstis:
Don't forget we are playing a level of pragmatism here - there's no way on
Earth I'm going to refactor the Guided Rule Editor for a 2 line fix to get
MVP'ed testable code. But the opinion here seems to be it'll be simple to
partially refactor to support unit testing.

porcelli:
"there's no way on Earth I'm going to refactor the Guided Rule Editor for a 2 line fix to get MVP'ed testable code" -> Why not?! :smile: 

this is not Guided Rule - that's considered `deprecated` and legacy... afaik this is not the case... implement proper test and MVP is the way to go unless there's a reason for not.

manstis:
I whole heartedly agree :-)

Salaboy:
@yurloc @porcelli @manstis :) MVP + test added , looking forward for some recommendations to improve those tests

manstis:
Nice tests - although it might have been nice to have one for a "success" state e.g. no errors, callback called and popup hidden.

A couple of comments:-

1) The asterisk in the i18n text makes me think there's a footnote somewhere. Is there?

2) You don't reset the view's errors each time "ok" is selected, so old errors that may have been corrected will remain highlighted as errors. I wasn't sure if such a scenario could exist, or whether there's only one validation?

guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/release/ReleaseScreenPopupPresenter.java
```diff
@@ -0,0 +1,101 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
yurloc: Missing license header.

guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/test/java/org/guvnor/asset/management/client/editors/project/structure/ReleaseRepositoryPresenterTest.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2014 JBoss Inc
```
yurloc: It's 2015 ;-)

Salaboy: Well it is a class that we wrote last year hehe

guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/release/ReleaseScreenPopupPresenter.java
Salaboy: I will add that

Salaboy:
@manstis good catches.. for 1) the asterisk in the constant is to express a wildcard, do you have any other suggestion? Basically it means whatever that start with the "release" string will match,  2) I've cleared up the error messages so if the user closes and open the popup again you don't see old errors. 

I don't mind adding a success test, I will do that just because I like you ;)

guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/test/java/org/guvnor/asset/management/client/editors/project/structure/ReleaseRepositoryPresenterTest.java
yurloc: Really? I thought it was the test you wrote just a couple of hours ago.

Salaboy: Hehe I might have copied from another class or copy+refactor another class, you are right for this one.

Salaboy: By the way, we should have a script for updating those headers somewhere..

psiroky: I am sure this is easily configurable in all major IDEs. You can define template and it will put the current year there when creating the class.

yurloc: In NetBeans IDE, I have the following template for Java Class files:
```
<#assign licenseFirst = "/*">
<#assign licensePrefix = " * ">
<#assign licenseLast = " */">
<#include "${project.licensePath}">
<#if package?? && package != "">
package ${package};

</#if>
public class ${name} {

}
```
The first four lines should also appear in templates for other file types like unit tests, enumerations, exceptions etc. Templates can be edited in Tools > Templates menu.

There are pre-configured license headers that can be viewed in project properties. If the project effective POM contains license element that matches one of the headers, it is used. Organization (the copyright owner) is also picked up from POM and year is generated correctly, too.

Salaboy:
suggested changes added and ready for review

manstis:
@Salaboy Hmmmm... this is a strange implementation of the MVP pattern ;)

Normally the P would contain a reference to the V; and Users would interact with the P; e.g. P.show(). Your implementation, I assume, requires the User to instantiate the V which contains a reference to the P.. I also assume Users would instantiate and then call ```configure(...)```. It might be an idea to move the call to ```clearWidgetsState()``` to an override ```show()``` rather than in ```configure(...)```. 

Have I misunderstood anything?

Salaboy:
@manstis that's probably correct, In fact I don't remember from where I copied these popups from, so it might be the case that originally copied from the wrong panel. I remember that our first implementations were @WorkbenchPresenters and then refactored to be view only + the configure method added.  Initially this was only a view, I've just separated the Presenter to be able to test that code easily. I didn't changed how the objects were instantiated mostly because all the other similar screens should be refactored and aligned as well.

Salaboy:
@manstis in fact what you are saying (as far as I understand) is that we need to change this one as well: RepositoryStructureViewImpl and RepositoryStructurePresenter to have the references to the presenters instead of the views. Is that a correct assumption?

guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/release/ReleaseScreenPopupPresenter.java
```diff
@@ -0,0 +1,111 @@
...
+                    return;
+                }
+
+                if ( isEmpty( view.getServerURLText().getText() ) ) {
+                    view.getServerURLTextGroup().setType( ControlGroupType.ERROR );
+                    view.getServerURLTextHelpInline().setText( Constants.INSTANCE.FieldMandatory0( "ServerURL" ) );
+
+                    return;
+                }
+
```
porcelli: @Salaboy this is not MVP - Presenter shouldn't have direct access to Widgets

manstis:
@Salaboy Nope; ```RepositoryStructurePresenter``` is quite different. There's nothing wrong with a V having access to a P; indeed when a V responds to, for example, a button click it is quite likely the V will call (for example) ```P.okButtonClicked()```. The P would then call a service etc. The P is instantiated by the User and a V created by the P passing, if necessary a reference to itself to the V. In your case the User is instantiating the V and the V creates an instance of the P.

So, in pseudo code, your MVP should be something like this:-

*User code*
```
ReleaseScreenPopupPresenter p = new ReleaseScreenPopupPresenter();
p.show();
```
*Presenter*
```
public class ReleaseScreenPopupPresenter {

    private ReleaseScreenPopupView view;

    public ReleaseScreenPopupPresenter() {
        this.view = new ReleaseScreenPopupViewImpl();
        this.view.setPresenter( this );
    }

    public void show() {
        this.view.show();
    }

    public void isValid( final String s ) {
        return !(s == null || s.isEmpty() );
    }

    public void doSomething() {
        //Do something
    }

}
````
*View interface*
````
public interface ReleaseScreenPopupView extends UberView<ReleaseScreenPopupPresenter> {

    void show();

}
````
*View implementation*
````
public class ReleaseScreenPopupViewImpl extends BaseModal implements ReleaseScreenPopupView {

    @UiField
    TextBox textBox;

    private ReleaseScreenPopupPresenter presenter;

    public ReleaseScreenPopupViewImpl() {
        //Setup widgets etc
    }

    @Override
    public void setPresenter( fina ReleaseScreenPopupPresenter presenter ) {
        this.presenter = presenter;
    }
    ...
    @UiHandler("button")
    void onButtonClick( final ClickEvent e ) {
        if( p.isValid( textBox.getText() ) {
            p.doSomething();
        }
    }

}
```

yurloc:
@Salaboy `CopyPoupView` interface, `CopyPopup` presenter and `CopyPopupViewImpl` in [uberfire-comons-editor](https://github.com/uberfire/uberfire-extensions/tree/087db6a447/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file) are an example of MVP pattern as described by Michael.

manstis:
@Salaboy and [tests](https://github.com/uberfire/uberfire-extensions/blob/087db6a447ae4f90a21b0df4453b9557e6c01447/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/file/CopyPopupTest.java) for the MVP pointed to by @yurloc

manstis:
@Salaboy oh, BTW, nobody is criticising anyone here. Please don't misinterpret comments!

Salaboy:
@manstis all cool.. I see a lot of changes for 6.3.x, but this is definitely the way to go.. each of my commits were based in the assumption that my changes should be limited to the scope of the BZ, but if we are now going all in for a proper MVP all the suggested changes make sense. Unfortunately I'm against the clock with this one, and I would like to make sure that this BZ is closed asap. I will submit one more commit for review and then I will be off.. hopefully someone else can pick it up if more changes are required.

Salaboy:
@manstis @porcelli  I've just pushed the refactoring, can you take a look to see if the injections are aligned with the rest of the MVPs? for some reason 6.3.x doesn't work in my local environment (kie-wb compiles all the permutations but the app doesn't start in hosted mode and wildfly).

manstis:
@Salaboy, IMO it's a great improvement, it's (more) aligned to MVP and has tests. Whilst we could critique this PR to death I think we should pragmatically accept the changes as a huge step forwards and merge.

So +1 from me.

Salaboy:
@manstis I really appreciate all your comments and your reviews. If you have more comments I would love to get this done perfectly so we can migrate all the other 3 screens that are extremely similar. So let's try to make this one as an example for further migrations. Unfortunately, I'm off for two weeks now and this is required for an urgent BZ, but please add as many notes as you want. Cheers

csadilek:
Since this it ready to merge the commits need to be squashed.

csadilek:
@Salaboy I've squashed the commits for you and pushed, enjoy your trip!

6.3.x: e9fff1e70f7cb2d6cabba918b68a61507c91b6bc

Separate PR for master is still required but can wait.

PR #179, opened 2015-11-04
Better arrow (U+2192) in Repository Structure title

yurloc:
The string "- >" that should represent the relation between repository and a project inside it looks awkward with the font currently used in the worbench. I have replaced it with a unicode arrow.

Before:
red hat jboss brms business central 2015-11-04 18-59-33

After:
red hat jboss brms business central 2015-11-04 19-01-05

### DIFF
--- a/guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/RepositoryStructurePresenter.java
+++ b/guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/RepositoryStructurePresenter.java
@@ -401,7 +401,7 @@ public class RepositoryStructurePresenter
 
             changeTitleWidgetEvent.fire( new ChangeTitleWidgetEvent(
                     placeRequest,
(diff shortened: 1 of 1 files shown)


manstis:
Am I wrong being concerned with the inclusion of the new arrow literal
rather than using /u2192 (sorry I forget the syntax)? Any foreseeable
problems with someone editing the file in the future with a non-unicode
editor (IDK vim for example?) Perhaps I'm too stuck on ASCII7 ;-)

manstis:
+1 to merge (even vim supports unicode)

yurloc:
Yes, I made the change using Vim :-) I agree it is safer to encode Unicode characters using the `\uXXXX` syntax. It can prevent an unwanted change with some kind of old-school/buggy editor but such a mistake could also be caught during the code review.

Just let me know if you want me to changed that.

manstis:
@yurloc I'm torn to be honest. Personally I prefer ```\uXXXX``` but it makes no logical reason in this modern (generally Unicode proof world); plus it seems really pedantic asking for it to be changed. So..

manstis:
+1 to merge.

csadilek:
Thanks!

master: ff4e2e0aaf2bb041b50059e0879b1487ca6674ea

PR #188, opened 2015-11-10
[GUVNOR-2279] Deploy artifact with encrypted password.

afcollins:
When deploying an artifact to a remote repository and the server
password is encrypted using master password, reuse the Authentication
that is built within the RepositorySystemSession.

GUVNOR-2279 log exceptions

GUVNOR-2279 - Test case.

GUVNOR-2279 - Cleaning test case.

GUVNOR-2279 - Resetting System properties after test

GUVNOR-2279 - Improve test case robustness.

Depending on runtime, settings.xml may not be found.  Resorting to
absolute file path to locate test resources.

Changed to mock Aether's static methods.

GUVNOR-2279 - Reintegrate test to backend module

Format changes, and move all test classes and resources to
guvnor-m2repo-backend project, rather than separate test project.


afcollins:
@manstis Squashed to a single commit.

manstis:
+1 to merge... thanks @afcollins!

manstis:
ok to test

psiroky:
ok to test

manstis:
@psiroky... why didn't it pick up my request?!?! Do I have a leading space (it didn't look like it in my reply email, but it does look like it above)

psiroky:
I am not sure. I was adding the comment to see if it picks it up or not. My initial suspect was the response by e-mail itself, if it somehow contains different info than regular comment. Those two comments look exactly the same and you definitely have the permission to approve the build as you in the previous PR for this issue.

manstis:
@psiroky ok, no problem. I'll comment on the website rather than by email in the future.

psiroky:
But this is definitely annoying. I will see if I can fix this somehow.

manstis:
+1 to merge

csadilek:
Thanks!

master: c0c2b4db2b5d8a582be8cc6d948a504520fac65b

afcollins:
Can I help backport to earlier branches?  I initially found this problem in BPMS6.1 GA.  Also, should I change JIRA issue status?

manstis:
We won't release older versions; but if you have a RedHat BRMS/BPMS
subscription you can request it's back ported to 6.x.

Alternatively if you want to build from source you can submit a PR for
whatever branch you need it in and I'll request it's merged.

Thanks for the fix by the way!

Cheers,

Mike

PR #190, opened 2015-11-11
webapp: use just single GWT module

psiroky:
* guvnor-webapp is used only for debugging/testing,
   so there really is no real need to have two different
   GWT modules for it (FastCompiled and Full). Having just one
   module should simplify the configuration and should also
   fix the SDM issue in IDEA introduced recently as the property
   gwt.user.agent.all was not properly replaced during the run

@manstis, @Rikkola please take a look. These are the same changes as for drools-wb in https://github.com/droolsjbpm/drools-wb/pull/88


manstis:
+1 to merge.

csadilek:
Thanks.

master: 9eb38bcf9eddce546d2a2630a1a9ebdc7c8793a7

PR #195, opened 2015-11-17
Adjust dashbuilder dependencies

dgutierr:
There was a missing dependency gwt-bootstrap3-extras. I've also simplified the dependency set.


manstis:
+1 to merge.

PR #197, opened 2015-11-19
Replace system output with logging

yurloc:
I find the messages about starting/ending asset management processes useful so I removed system output and raised level of the logging. I also removed date parameter from the message as log messages usually have time stamps.


manstis:
+1 to merge.

csadilek:
Thanks!

master: 8d2f008209cdaab396eb028e98b290ece728ac82

PR #202, opened 2015-11-25
UF-244: (Work-around) Default DEV environment to Wildfly 10.0.0(CR4).…

manstis:
See https://issues.jboss.org/browse/UF-244

This PR provides a workaround to the issue (of ```Async``` EJBs not running on Wildfly 8.1 - which is our Hosted Mode environment). Support for System Property ```org.uberfire.async.executor.safemode``` to force use of ```Executors.newCachedThreadPool``` has been added.

I have verified both "indexing" and "incremental builds" work in Hosted Mode when the ```org.uberfire.async.executor.safemode``` System Property is set to ```true``` in (IDEAs) Launch Configuration.

### DIFF
--- a/guvnor-project/guvnor-project-builder/src/main/java/org/guvnor/common/services/builder/IncrementalBuilderExecutorManagerFactoryImpl.java
+++ b/guvnor-project/guvnor-project-builder/src/main/java/org/guvnor/common/services/builder/IncrementalBuilderExecutorManagerFactoryImpl.java
@@ -25,9 +25,15 @@ import org.guvnor.common.services.project.builder.model.IncrementalBuildResults;
 import org.guvnor.common.services.project.builder.service.BuildService;
 import org.guvnor.common.services.project.model.Project;
 import org.guvnor.common.services.project.service.ProjectService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


manstis:
See https://github.com/uberfire/uberfire/pull/230

manstis:
@mswiderski @porcelli @csadilek @wmedvede Please review.

mswiderski:
+1 from me, I guess it would make sense to add system property directly to extraJvmArgs [1] (at least kie-wb and kie-drools-wb) of gwt-maven-plugin so those who run it with mvn gwt:run can benefit from this change too

[1]https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-webapp/pom.xml#L1515

manstis:
@mswiderski Good idea.. I'll do another PR for kie-wb-distributions (I'd be emailing BSIG about it too; so others can add the System Property to their respective "showcases" if needed; e.g. drools-wb-webapp, designer-standalone, jbpm-console-standalon etc).

manstis:
See https://github.com/droolsjbpm/kie-wb-distributions/pull/154

csadilek:
master: 30cd38003d4c8c4870d8086ff2e8e0e46dda38ce

PR #204, opened 2015-12-02
GUVNOR-2278 : Project Settings: Dependencies should expose dependency scope

Rikkola:
The goal was to add make it possible to edit scope of the dependencies defined in pom.xml, but while I was there...

The reasons for the extra changes:
* Any pom.xml we open should be saved as it was before. Project editor just allows the user to edit a selected set of fields. We do not want to overwrite anything that the user might have hard edited in with a text editor, by passing our guided editor for the pom.xml. 
* Guvnor should not add the kie maven plugin. This leaks drools/jbpm into guvnor and guvnor should be reusable.
* @manstis already did some changes for the NewProjectEditor, but I changed the passing of a GAV to POM. The wizard still edits just the GAV part, but does not care if there are plugins or parent GAV declared for the to be new project.
* If I touched something that was lacking tests I wrote some tests.

https://github.com/droolsjbpm/guvnor/pull/204
https://github.com/droolsjbpm/kie-wb-common/pull/171
https://github.com/droolsjbpm/drools-wb/pull/97


guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/model/Dependencies.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+package org.guvnor.common.services.project.model;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+
+public class Dependencies
```
manstis: Should this be Portable too?

Rikkola: For some reason it works as it is, but it should yes. Making it Portable.

guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/BuildContentHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+            if ( from.getPlugins() != null ) {
+                to.setPlugins( new PluginUpdater( to.getPlugins() ).update( from.getPlugins() ) );
+            }
+
+            return to;
+        }
+    }
+
+    private Plugin update( org.guvnor.common.services.project.model.Plugin from,
```
manstis: Is this ever used? (I might be going blind - or was it before ```PluginUpdater``` was used?)

guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/PluginUpdater.java
```diff
@@ -0,0 +1,38 @@
...
+public class PluginUpdater {
+
+    private final List<Plugin> plugins;
+
+    public PluginUpdater( List<Plugin> plugins ) {
+        this.plugins = plugins;
+    }
+
+    public List<Plugin> update( ArrayList<org.guvnor.common.services.project.model.Plugin> from ) {
+
```
manstis: Nice to see you've left some room for future improvements ;)

guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/BuildContentHandler.java
Rikkola: Not used, left over from moving the functionality to another class. Removing it.

guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/PluginUpdater.java
Rikkola: Now that I took a look. The method above should be here. Adding test coverage and a fix.

Rikkola:
@manstis changes are done based on comments.

manstis:
+1 to merge.

csadilek:
Thanks.

master: 976e7611c8f809a78b52f1b24f147180bdcf3642

PR #205, opened 2015-12-07
GUVNOR-2317: Wrong Commit History layout for long commit messages

manstis:
See https://issues.jboss.org/browse/GUVNOR-2317

Tidied up separation between M, V and P; fixed the CSS issues and wrote some tests.

### DIFF
--- a/guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/edit/RepositoryEditorView.java
+++ b/guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/edit/RepositoryEditorView.java
@@ -28,9 +28,8 @@ import com.google.gwt.uibinder.client.UiField;
 import com.google.gwt.user.client.Element;
 import com.google.gwt.user.client.ui.Composite;
 import com.google.gwt.user.client.ui.FlowPanel;
-import com.google.gwt.user.client.ui.RequiresResize;
 import com.google.gwt.user.client.ui.Widget;
-import org.guvnor.structure.client.navigator.CommitNavigator;
+import org.guvnor.structure.client.navigator.CommitNavigatorEntry;
 import org.guvnor.structure.client.resources.i18n.CommonConstants;
 import org.guvnor.structure.repositories.PublicURI;
 import org.gwtbootstrap3.client.ui.Button;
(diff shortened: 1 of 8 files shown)


guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/edit/RepositoryEditorView.java
```diff
@@ -145,9 +132,7 @@ public void execute( final VersionRecord record ) {
         loadMore.addClickHandler( new ClickHandler() {
             @Override
             public void onClick( ClickEvent event ) {
-                if ( commitNavigator != null ) {
-                    presenter.getLoadMoreHistory( commitNavigator.getLastIndex() );
-                }
+                presenter.getLoadMoreHistory( history.getWidgetCount() );
```
Rikkola: I have used the presenter.on*() format for any action that is done to presenter from the view impl. For example a refresh button click would be presenter.onRefresh(), not caring about what the actual widget is, but on what the action is. 
Just a thing I do to try and keep the readability better for the presenter methods and I feel it also helps reading the unit tests. 
Just a proposal for what I think we should do. Now that I think these methods should probably also be package private.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/navigator/CommitNavigatorEntry.java
```diff
@@ -36,87 +37,46 @@
...
+    }
 
-        navigator.setWidget( row, col, new Widget() {{
-            setElement( messageCol );
-        }} );
+    private void initWidget( final VersionRecord versionRecord,
+                             final ParameterizedCommand<VersionRecord> onRevertCommand ) {
+        message.getElement().setInnerSafeHtml( makeSafeHtml( versionRecord.comment() ) );
+        author.getElement().setInnerSafeHtml( makeSafeHtml( versionRecord.author() + " - " ) );
+        date.getElement().setInnerSafeHtml( makeSafeHtml( fmt.format( versionRecord.date() ) ) );
```
Rikkola: Way better to use UiBinder here, but would it be possible to use a HTML element or just a Label in the ui.xml for these? 
There might be a reason for doing it this way and I am missing it. Just feels more complex and if I just look at the ui.xml it is not clear what is shown there.

Rikkola:
+1

csadilek:
Thanks.

master: 664e3743df7995e2abfcb8ad54a49b20eace04d3

PR #214, opened 2016-01-08
BZ1296498: [GSS] (6.2.z) Opening a repository takes a long time in BRMS 6.2

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1296498

The issue was introduced by BPMSPL-100 which added support for "tag filtering". This PR makes retrieval of tag meta-data optional only when "tag filtering" is enabled in the UI.


manstis:
See https://github.com/droolsjbpm/kie-wb-common/pull/192 too.

porcelli:
https://github.com/droolsjbpm/guvnor/commit/2ee16377e90c2f8b66f8452c3ed7bfe2e986d2e0

PR #220, opened 2016-01-21
GUVNOR-2345 : Remove test setup redundancy

Rikkola:
Trying to unify the way we do server side testing, reduce code redundancy in tests and reduce the time we spend setting up the tests.

This breaks kie-wb-common if the PR in kie-wb-common is not in.
https://github.com/droolsjbpm/kie-wb-common/pull/206


manstis:
+1 to merge.

csadilek:
master: a3d9abe43122b18c5170bc5e37d6c2e8ef31946b

PR #226, opened 2016-01-28
Minor UI styling updates

cristianonicolai:
Reviewed forms according to PatternFly guidelines
Included Bootstrap grid to allow better spacing between screens


manstis:
+1 to merge (it's only UI composition and separation...)

Technically, in a totalitarian society, *master* should use Errai UI for screens.. however since this is really only a tweak of what already exists blah, blah, blah I think we'll be OK :)

csadilek:
master: 308d995fa87a9f4c8ad7fcecf40e900ff828c31a

PR #228, opened 2016-01-28
User with no privileges for repository can view and modify assets in that repository

ederign:


psiroky:
The test failures are not related and are kind of random. I've seen them fail in past as well. We need to take a look at those and fix them once and for all.

manstis:
Licenses for new files show 2015; but otherwise fine... +1 to merge.

csadilek:
6.4.x: 21e76547a0d7efb10de67184c6bb287d64aaf6dc

PR #232, opened 2016-01-29
Fix dep cycle in BZ1289724

ederign:


manstis:
Won't @Inject ProjectService<? extends Project> throw a WELD unsatisfied injection point if there is no implementation available (not the case for KIE, but technically possible). @Any might have been less likely to potentially lead to WELD errors...

ederign:
@manstis should I use      private Instance< ProjectService<? extends Project> > projectService2 ?

And how about this? https://github.com/ederign/guvnor/blob/c067f5218397430cc487e134e12baeae36625757/guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureServiceImpl.java#L65

guvnor-inbox/guvnor-inbox-backend/src/main/java/org/guvnor/inbox/backend/server/security/InboxEntrySecurity.java
```diff
@@ -107,6 +105,9 @@ private boolean canAccessRepository( Set<Repository> authorizedRepositories, Rep
     }
 
     Project getInboxEntryProject( final InboxEntry inboxEntry ) {
+        if ( projectService == null ) {
```
porcelli: why?

guvnor-inbox/guvnor-inbox-backend/src/test/java/org/guvnor/inbox/backend/server/security/InboxEntrySecurityTest.java
```diff
@@ -24,10 +26,10 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.kie.workbench.common.services.shared.project.KieProjectService;
 import org.mockito.runners.MockitoJUnitRunner;
 import org.uberfire.security.authz.AuthorizationManager;
 
+import javax.enterprise.inject.Instance;
```
porcelli: ?

manstis:
@ederign hehe, well that's a candidate for failure too ;-) I guess your PR is just fine (or as good as what we gave elsewhere!) In reality the scenario is probably unlikely so....

manstis:
+1 to merge.

manstis:
@ederign uh oh... someone's woken the sleeping @porcelli beast.. easy-times will come to an end now ;-)

ederign:
@porcelli  I will update it.

ederign:
and @manstis  thanks for the review.

ederign:
6.4: https://github.com/droolsjbpm/guvnor/pull/233

ederign:
6.3: https://github.com/droolsjbpm/guvnor/pull/234

manstis:
Failing test is unrelated.

+1 to merge.

manstis:
@psiroky Any ideas about the failing test?!?

csadilek:
Yes, test failure unrelated but persistent. @psiroky I can help with this on Monday, unless you already have an idea where the invalid bytecode comes from. It wasn't a problem last night after upgrading.

csadilek:
master: b88dd1139ac1fbf5339f9f8ebcc725d566044cd1

psiroky:
I'll take a look a this. But from the quick scroll through the test output, I currently have no idea what is going on.

csadilek:
@psiroky @manstis @ederign I got to the bottom of this now. It's caused by powermock using the wrong version of javassist. We have two versions of javassist on the classpath (javassist:javassist 3.12.1.GA brought in by resteasy-cdi-jar and org.javassist(different groupId):javassist 3.18.2-GA brought in by uberfire-api). Using the 3.18.2-GA or newer will fix the issue. 

The easiest fix is to add the following before the guvnor-service-api dependency:

```
<dependency>
  <groupId>org.javassist</groupId>
  <artifactId>javassist</artifactId>
  <scope>test</scope>
</dependency>
```

@psiroky does that sound ok to you?

psiroky:
@csadilek thanks for looking into this! I think we need to exclude the wrong javassist (javassist:javassist) and override it with the good one (org.javassist:javassist). Otherwise anything depending on that guvnor module can easily get the wrong one transitively as well. 

I am even thinking about banning the javassist:javassist using enforcer since this is no the first time we are having these kind of issues.

psiroky:
The root cause is in `errai-security-server`. I am preparing a PR to fix it directly there.

csadilek:
@psiroky yes agreed, that would be the best solution. resteasy-cdi seems to be fine with the exclude. Of course, we might have missed something...

csadilek:
@psiroky we can then also get rid of the exclusions we added in other poms: https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-webapp/pom.xml#L140

psiroky:
+1 for removing those exclusions.

PR #244, opened 2016-02-11
GUVNOR-2384: Artifact Repository and Artifacts Modal never finish loading

manstis:
See https://issues.jboss.org/browse/GUVNOR-2384

This was unfortunately introduced with the fix for https://issues.jboss.org/browse/GUVNOR-2373

Which goes to prove, even with Unit Tests, there's no substitute for actually running things in the UI!!


csadilek:
+1

csadilek:
master: f6f1ed53092de1aff32fdb0f1aeedc1b4b09869a

PR #245, opened 2016-02-11
[6.4.x] GUVNOR-2384: Artifact Repository and Artifacts Modal never finish loading

manstis:
See https://issues.jboss.org/browse/GUVNOR-2384

This was unfortunately introduced with the fix for https://issues.jboss.org/browse/GUVNOR-2373

Which goes to prove, even with Unit Tests, there's no substitute for actually running things in the UI!!

(cherry picked from commit 5cab265d0513f3de569a7daead22e287200fdefb)


csadilek:
6.4.x: 73a710eab3bfb463549e9472c7b09a376c97e80b

PR #248, opened 2016-02-16
GUVNOR-2365: Repository remains in repositories list after it is deleted

paulovmr:


guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/list/RepositoriesView.java
```diff
@@ -86,11 +87,14 @@ public boolean confirmDeleteRepository( final Repository repository ) {
 
     @Override
     public void removeIfExists( final Repository repository ) {
-        Widget w = repositoryToWidgetMap.remove( repository );
-        if ( w == null ) {
-            return;
+        Iterator<Map.Entry<Repository, Widget>> it = repositoryToWidgetMap.entrySet().iterator();
```
manstis: Does this fix imply the ```GitRepository.equals()``` and ```GitRepository.hashCode()``` methods are broken? Whilst fixing these would arguably be the more correct "technical" solution it also carries the higher risk of breaking something elsewhere.. so happy to fix the JIRA as you have done.

manstis:
+1 to merge.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/list/RepositoriesView.java
paulovmr: @manstis After some debugging, that is the only possibility that I came up with too, but I haven't been able to see the motive why they are broken.

PR #250, opened 2016-02-17
BZ1078112 - Move upload and refresh buttons into toolbar

cristianonicolai:
Depends on https://github.com/uberfire/uberfire-extensions/pull/211

### DIFF
--- a/guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/test/java/org/guvnor/m2repo/client/M2RepoEditorPresenterTest.java
+++ b/guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/test/java/org/guvnor/m2repo/client/M2RepoEditorPresenterTest.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/test/java/org/guvnor/m2repo/client/M2RepoEditorPresenterTest.java
```diff
@@ -0,0 +1,90 @@
...
+            }
+        });
+
+        final MenuCustom refresh = (MenuCustom) presenter.getMenus().getItems().get(1);
+        refresh.build();
+
+        clickHandler.onClick(new ClickEvent() {
+        });
+
+        verify(refreshEvents).fire(any(M2RepoRefreshEvent.class));
```
manstis: Out of interest.. what do I do wrong?!?!

Whenever I use ```EventSourceMock``` I get ```UnsupportedOperationException``` errors (due to the ```fire()``` implementation) and normally have to override it to not throw the exception. I see you don't override and you also expect ```fire()``` to be have been called... so either I'm doing something wrong (and happy to be corrected) or this test fails?

manstis:
+1 to merge (test clearly passes.... and I wait to be educated)

guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/test/java/org/guvnor/m2repo/client/M2RepoEditorPresenterTest.java
cristianonicolai: @manstis I guess I just did something different to achieve the same thing. Because I'm using `@Spy` for `refreshEvents`, it allows me to partially overwrite the implementation using `doNothing().when(refreshEvents).fire(any(M2RepoRefreshEvent.class));`, that way I get away from that code throwing an exception.

manstis: @cristianonicolai Ah yes.. I see the ```setup()``` now.. mystery solved!

psiroky:
Merged into 6.4.x

PR #258, opened 2016-02-22
BZ-1305056 - Unable to automatically create further branches in Manag…

mswiderski:
…ed Git Repository for BRMS 6.2

makes sure that only workbench instance that requested asset mgmt process to be executed will be the one that actually executes it. There is no way to invoke actual process from within guvnor so just added tests to make sure proper parameters are there.


manstis:
@mrietveld Do we have any integration tests in ```kie-wb-distributions``` to support "Asset Management" process testing? It seems to be a bit of  black hole.

manstis:
+1 to merge -- but this does highlight a hole in test coverage (for "Asset Management").

mswiderski:
@manstis indeed with have a gap here and mainly because of dependencies as guvnor cannot depend on jbpm thus this mechanism of sending cdi events to actually trigger processes

mrietveld:
@manstis Nope, not that i know of (Asset Management process integration tests).

PR #267, opened 2016-03-01
JBPM-4863 - Allow to make managed repo while cloning

mswiderski:
as the jira name suggests it allows to mark as managed repository when cloning. It is restricted to roles for asset mgmt to be consistent with managed repos in general.

@manstis @wmedvede do you mind taking a look?


manstis:
+1 to merge, seems good to me!

csadilek:
master: c3f6ea54dd05b39040a1bd7dccf3c0f085baba86

PR #272, opened 2016-03-04
GUVNOR-2452: general improvements on artifact list in order to be eas…

porcelli:
…y to reuse (needed by new exec server ui)

Now bug is fixed, ready to review.

### DIFF
--- a/guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/editor/MavenRepositoryPagedJarTable.java
+++ b/guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/editor/MavenRepositoryPagedJarTable.java
@@ -15,7 +15,6 @@
 
 package org.guvnor.m2repo.client.editor;
 
-import javax.annotation.PostConstruct;
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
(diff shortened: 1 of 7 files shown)


guvnor-m2repo-editor/guvnor-m2repo-editor-client/src/main/java/org/guvnor/m2repo/client/editor/MavenRepositoryPagedJarTable.java
```diff
@@ -39,84 +41,95 @@
...
+    public void search( String filter ) {
+        presenter().search( filter );
+    }
+
+    public void refresh() {
+        presenter().refresh();
+    }
+
+    private ArtifactListPresenter presenter() {
+        if ( presenter == null ) {
```
manstis: There must be a good explanation, but this suggests the View is instantiated without a Presenter; normally we'd work the other way around - Presenter first then View... what's the reasoning here?

porcelli: This is not the view's presenter.. this is the ArtifactListPresenter (widget) presenter. And this is related to @AfterInitialization problem that is workarounded here with a lazy init.

porcelli: I'd say this implementation is far from ideal... it should be refactored; but as this is supposed to be last 6.x release I'd prefer to make the change as minimium as possible.

manstis:
+1 to merge.

csadilek:
6.4.x: bc63bea598830a02c3ae654c2c577aded6a07e30

PR #274, opened 2016-03-07
BZ#1315341: Registered the right event handler

jomarko:


jomarko:
Wrong event was fired directly by gwtboostrap3, this is fixed by [1] but I do not if it is assumed to use newer version of gwtboostrap3.

[1]
https://github.com/gwtbootstrap3/gwtbootstrap3/commit/b2d121de3a5546b128c963c44ba197445fdf557e

manstis:
Same as my comment on the other PR (pasted here for completeness):

@jomarko Hi.. I assume this PR is dependent on the GWT Boostrap fix? We wouldn't automagically upgrade the version of GWT Bootstrap we use - unless you want to include in a PR? I also wonder whether the hack for https://issues.jboss.org/browse/GUVNOR-2302 (that is referenced by the GWT Bootstrap [issue](https://github.com/gwtbootstrap3/gwtbootstrap3/issues/375) should also be removed - assuming we can?!?)

jomarko:
Ready for review

manstis:
@jomarko Thanks.. +1 to merge.

csadilek:
master: 4172a6b326912ba83976046d3b2c4cc6a3e8b93c

PR #288, opened 2016-03-15
Fix build after PageTable refactoring

csadilek:
@ederign or @manstis see source...one more call site that was missed in the latest refactoring.


manstis:
+1 to merge.

csadilek:
master: 48b529693373e74cc3b5d386c39941ffc3d41410

PR #292, opened 2016-03-19
GUVNOR-2352 : Adding of manual dependency does not work correctly

Rikkola:
Removed the packages from `Dependency`. They were lazy loaded by sending the instances to `DependencyService` for the names to be filled up. This can cause bugs and `Dependency` class is really our version of the Maven Dependency. The version Maven has does not have the info about packages either.

https://github.com/droolsjbpm/kie-wb-common/pull/321


manstis:
+1 to merge.

csadilek:
master: 7e155b15ae97d567a53ac832f99748482fac7bc0

PR #296, opened 2016-03-22
BZ-1315671: dev/release branches initially not listed in Repository Editor after managed repository created

wmedvede:


wmedvede:
@manstis 

Hello Mike, this is the PR we chatted about. NOT for merging yet.
Could you please take a look?

Thanks.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/list/RepositoryItemPresenter.java
```diff
@@ -71,6 +71,11 @@ public void setRepository( final Repository repository,
         view.refresh();
     }
 
+    public void refreshBranches() {
+        populateBranches( view.getSelectedBranch() );
+        view.refresh();
```
manstis: Minor comment; ```setSelectedBranch(..)``` in ```populateBranches(..)``` calls ```branchesDropdown.refresh()``` which is the same as ```view.refresh()``` so the call here should not be strictly needed; but since it's not an expensive operation it probably does no harm.

manstis:
@wmedvede It looks fine to me (I assume you'll add a couple of tests)?

wmedvede:
@manstis Ok, thanks! Tests are comming up with the mergeable version.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/list/RepositoryItemPresenter.java
wmedvede: I would prefer keep this invocation for this case, since the Select component is sometimes tricky.

wmedvede:
@manstis Hello Mike, the tests has been added. Ready to merge in my opinion.

manstis:
Thanks @wmedvede 

+1 to merge.

csadilek:
6.4.x: f13ebdc190eb49d20fae9ae9665b60ed5611df96

PR #302, opened 2016-04-13
[6.4.x] Bug 1324596 - [GSS] [6.2.z] 6.2.2 unable to open/edit some assets with specific user

Rikkola:
https://bugzilla.redhat.com/show_bug.cgi?id=1324592

This is now a release blocker.

Targeting 6.4.x first, since couldn't get the super dev mode to work in master.

### DIFF
--- a/guvnor-inbox/guvnor-inbox-backend/src/main/java/org/guvnor/inbox/backend/server/security/InboxEntrySecurity.java
+++ b/guvnor-inbox/guvnor-inbox-backend/src/main/java/org/guvnor/inbox/backend/server/security/InboxEntrySecurity.java
@@ -15,6 +15,14 @@
  */
 package org.guvnor.inbox.backend.server.security;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
 import org.guvnor.common.services.project.model.Project;
 import org.guvnor.common.services.project.service.ProjectService;
 import org.guvnor.inbox.backend.server.InboxEntry;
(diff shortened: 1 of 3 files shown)


manstis:
Looks good.. but what consumes the "Repository" (that could now be null.. is it null safe.. or are we likely to get a NPE somewhere now)?

guvnor-inbox/guvnor-inbox-backend/src/main/java/org/guvnor/inbox/backend/server/security/InboxEntrySecurity.java
```diff
@@ -114,9 +119,15 @@ private boolean thereIsNoRepositoryAssociated( Repository inboxEntryRepository )
...
 
     Repository getInboxEntryRepository( InboxEntry inboxEntry ) {
-        final Path path = Paths.get( inboxEntry.getItemPath() );
-        final FileSystem fileSystem = path.getFileSystem();
-        return configuredRepositories.getRepositoryByRepositoryFileSystem( fileSystem );
+        try {
+
+            final Path path = Paths.get( inboxEntry.getItemPath() );
+            final FileSystem fileSystem = path.getFileSystem();
+            return configuredRepositories.getRepositoryByRepositoryFileSystem( fileSystem );
```
Rikkola: @manstis this can also return null. So we are safe.

Rikkola:
@manstis and here is a check for null repository. I also saw a test for the situation.

https://github.com/Rikkola/guvnor/blob/BZ-1324596-deleted-repository-breaks-editors/guvnor-inbox/guvnor-inbox-backend/src/main/java/org/guvnor/inbox/backend/server/security/InboxEntrySecurity.java#L85

Rikkola:
Here is the test for situations when the repository is null.
https://github.com/droolsjbpm/guvnor/blob/23794cffc442d1f13e0eb12430b4bbb6d39f3916/guvnor-inbox/guvnor-inbox-backend/src/test/java/org/guvnor/inbox/backend/server/security/InboxEntrySecurityTest.java#L79-79

Rikkola:
Then there is the question if it should be marked null, most likely not, but we have no way to know about deleted repositories ATM. Now if you delete a repo you lose all information about it. Works differently than removing files or packages, where you always have the file/folder in version history.

manstis:
OK, thanks @Rikkola +1 to merge (as and when the BZ gets blocker+ approval)

krisv:
+1, since this can freeze entire workbench, this is considered a blocker for the release, please go ahead and merge this in 6.4.x

PR #311, opened 2016-04-20
MetadataCreatorTest doesn't compile on jdk 1.6

baldimir:
- Fixed MetadataCreatorTest compilation issues on jdk 1.6.


baldimir:
Hi, I didn't want to create JIRA for this, because it's only a fix in test. The problem is that when trying to run Guvnor tests on jdk 1.6, this test doesn't compile. JDK 1.6 is still supported configuration.

psiroky:
Yeah, we do not enforce JDK 6 compatibility for tests. Just for the `src/main` stuff. This is however very trivial _test_ change, so +1 once the build succeeds.

PR #314, opened 2016-05-11
GUVNOR-2486 : Project structure screen is not refreshed properly after changes in branch selection

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2486

https://github.com/droolsjbpm/kie-wb-common/pull/356

### DIFF
--- a/guvnor-asset-mgmt/guvnor-asset-mgmt-api/src/main/java/org/guvnor/asset/management/model/RepositoryStructureModel.java
+++ b/guvnor-asset-mgmt/guvnor-asset-mgmt-api/src/main/java/org/guvnor/asset/management/model/RepositoryStructureModel.java
@@ -16,6 +16,7 @@
 package org.guvnor.asset.management.model;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -34,8 +35,6 @@ public class RepositoryStructureModel {
     private Metadata POMMetaData;
 
     private Path pathToPOM;
-
-    private Path pathToImports;
 
     private List<String> modules = new ArrayList<String>( );
 
@@ -70,10 +69,6 @@ public class RepositoryStructureModel {
         return pathToPOM;
     }
 
-    public Path getPathToImports() {
-        return pathToImports;
-    }
-
     public void setPathToPOM( Path pathToPOM ) {
         this.pathToPOM = pathToPOM;
     }
@@ -82,8 +77,11 @@ public class RepositoryStructureModel {
         return modules;
     }
 
-    public void setModules( List<String> modules ) {
-        this.modules = modules;
+    public void setModules( final Collection<String> modules ) {
+        this.modules.clear();
+        for ( final String module : modules ) {
+            this.modules.add( module );
+        }
     }
 
     public boolean isMultiModule() {
--- a/guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureModelLoader.java
+++ b/guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureModelLoader.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.asset.management.backend.service;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import javax.inject.Inject;
+
+import org.guvnor.asset.management.model.RepositoryStructureModel;
+import org.guvnor.common.services.project.model.POM;
+import org.guvnor.common.services.project.model.Project;
+import org.guvnor.common.services.project.service.POMService;
+import org.guvnor.common.services.project.service.ProjectService;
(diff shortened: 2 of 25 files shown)


Rikkola:
Builds pass, the PRs are ready for review.

guvnor-asset-mgmt/guvnor-asset-mgmt-api/src/main/java/org/guvnor/asset/management/model/RepositoryStructureModel.java
```diff
@@ -82,8 +77,10 @@ public void setPathToPOM( Path pathToPOM ) {
         return modules;
     }
 
-    public void setModules( List<String> modules ) {
-        this.modules = modules;
+    public void setModules( final Collection<String> modules ) {
+        for ( final String module : modules ) {
```
manstis: Since you're now populating the existing ```List``` should we call ```this.modules.clear()```?

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureModelLoader.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        if ( managedStatus != null ) {
+            model.setManaged( managedStatus );
+        }
+
+        model.setOrphanProjects( new ArrayList<>( repositoryProjects ) );
+        for ( Project orphanProject : repositoryProjects ) {
+            final POM pom = pomService.load( orphanProject.getPomXMLPath() );
+            model.getOrphanProjectsPOM().put( orphanProject.getSignatureId(),
```
manstis: Is ```GAV``` not a suitable signature?

guvnor-asset-mgmt/guvnor-asset-mgmt-api/src/main/java/org/guvnor/asset/management/model/RepositoryStructureModel.java
Rikkola: We probably should. Just in case.

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureModelLoader.java
Rikkola: Need to check the logic behind this.

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureModelLoader.java
```diff
@@ -0,0 +1,155 @@
...
+
+        for ( final String moduleName : moduleNames ) {
+            result.put( moduleName,
+                        projectService.resolveProject( Paths.convert( parentPath.resolve( moduleName ) ) ) );
+        }
+
+        return result;
+    }
+
+    private Boolean getManagedStatus( final Repository repository ) {
```
manstis: Perhaps an ```enum``` for the tri-state would be better? MANAGED, UNMANAGED, UNKNOWN?

guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/RepositoryStructurePresenter.java
```diff
@@ -187,6 +188,8 @@ public RepositoryStructurePresenter( final RepositoryStructureView view,
         this.workbenchContext = workbenchContext;
         this.wizard = wizard;
 
+        workbenchContext.addChangeHandler( this );
```
manstis: And so we move away from CDI events.... ;)

manstis:
A few minor comments (some of which existed in the original code, before your refactors).

Rikkola:
I'll look into them. Best to check and fix when we can.

guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/RepositoryStructurePresenter.java
Rikkola: I think they can be used when branch has been created or a file opened, but when it is possible to just use a change listener it is a lot safer and easier to take that path. Here we are using the context as a central location for the current, well, context. So instead of adding the Observers for each event type that changes the context in each location that follows context changes, this is just a lot safer way.

The problems with the old approach were:
- We had events both notifying about a context change and making a context change. Due to bad naming perhaps, but it was hard to see what event A should be used for. 
- We had events updating the context, even when there was no actual server side changes. Context and what was on the file system became out of sync. Direction of events and change listeners should be server -> context -> anything that cares about context
- Context always knows what is the current state. Previously the classes listening to context listened to the same events as context. Now who knows what picks up the event first and then the screens we have are out of sync

I know you @manstis probably know about most of this. Wrote the above in case anyone alse is listening. I think we should make this a coding pattern for context, kind of like the EventBus that is used for a single editor instance. So instead of going wild with global events we narrow down some areas where event listener or scoped event bus is better.

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureModelLoader.java
Rikkola: For all I can think of, is that this signature id works even when the GAV is incomplete or not yet set. The signature id is the project class type, in our case always KieProject + the path to project root. Feels to me like the project root would be good enough.

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureModelLoader.java
Rikkola: Good catch, the Boolean indeed has 3 states.

Rikkola:
Fixed the ones I dared. Not sure about getOrphanProjectsPOM() at this point, we could change it to project root path or leave it as it is for now.

manstis:
+1 to merge.

PR #313, opened 2016-05-10
Remove unused guvnor-webapp

psiroky:

PR #315, opened 2016-05-11
[6.4.x] remove unused guvnor-webapp

psiroky:
Backport of #313. The webapp is not used later on (in downstream repository) so it safe to remove.


manstis:
+1 to merge.

PR #316, opened 2016-05-11
[6.3.x] MetadataCreatorTest doesn't compile on jdk 1.6 - fixed compilation is…

psiroky:
…sues.

Only test fixes.


manstis:
+1 to merge.

PR #319, opened 2016-05-26
GUVNOR-2522 : NPE after entering authoring perspective / creating new project

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2522

Related:
https://github.com/droolsjbpm/kie-wb-common/pull/369

While fixing the actual bug. I had doubts about RepositoryStructureServices.load(Repository,Branch) and branch not always being set up. I tried to change the code so that all the possible scenarios where load is used are safe.

While there RepositoryConfigurationPresenter had quite a lot in it. So moved few parts to new classes.


Rikkola:
Guvnor changes could be have a different ticket if that is preferred?

manstis:
@Rikkola If it relates to the same issue; then the same JIRA number is fine.

Rikkola:
Ok, I made a new Jira ticket just in case. Linking it to the related tickets
https://issues.jboss.org/browse/GUVNOR-2531

Rikkola:
Need to check how tests go. I skipped kie-wb-distribution build locally. That might fail.

Other than that the PRs are ready for review.

manstis:
+1 to merge. @Rikkola ```kie-wb-distributions``` appears unaffected! :+1:

PR #320, opened 2016-05-30
JBPM-5030: Access control based on the new UF's authz subsystem

dgutierr:
Changes required to support the new UF's security APIs as well as to integrate the following resource types intro the permission editor: Org. Units, Repositories & Projects.

This PR depends on uberfire/uberfire#321


guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/service/ProjectSearchService.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Collection;
+
+import org.guvnor.common.services.project.model.Project;
+import org.jboss.errai.bus.server.annotations.Remote;
+
+/**
+ * {@link Project} remote search interface
+ */
+@Remote
+public interface ProjectSearchService {
```
porcelli: still not completely confident about this search service; search for global projects sounds like too much (think abotu hundreds of projects in a large organization)... gut feeling is that this search would need some context.. and with a context would be possible to load everything into a data structure and send it to client to handle

guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/ProjectSearchServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+        return results;
+    }
+
+    @Override
+    public Collection<Project> searchById(Collection<String> ids) {
+        List<Project> results = new ArrayList<>();
+        for (Repository repository : repositoryService.getRepositories()) {
+            ProjectService projectService = getProjectService();
+            if (projectService != null) {
+                Set<Project> repositoryProjects = projectService.getProjects(repository, "master");
```
porcelli: iirc, this is a cost operation

guvnor-structure/guvnor-structure-api/src/main/java/org/guvnor/structure/organizationalunit/OrganizationalUnitSearchService.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Collection;
+
+import org.jboss.errai.bus.server.annotations.Remote;
+
+/**
+ * {@link OrganizationalUnit} remote search interface
+ */
+@Remote
+public interface OrganizationalUnitSearchService {
```
porcelli: too much to use a search service... maybe even all github org units could be loaded directly into a data structure and handled by client side :P

guvnor-structure/guvnor-structure-api/src/main/java/org/guvnor/structure/repositories/RepositorySearchService.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Collection;
+
+import org.jboss.errai.bus.server.annotations.Remote;
+
+/**
+ * {@link Repository} remote search interface
+ */
+@Remote
+public interface RepositorySearchService {
```
porcelli: similar thought: sounds like too much, probably could be all loaded to client with some context

porcelli:
some concerns... but would be ok to merge: +1

guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/service/ProjectSearchService.java
dgutierr: I totally agree.

This is an initial implementation that works fairly well. At least for an small number of instances.

AFAIK @mrietveld is working on index based implementations for searching OUs, repos & projects. He plans to apply this new implementation once this PR is merged on master.

Aside from that, I also agree some context is required, like the repo user wants to manipulate its permissions, so that only that repo's project are fetched. Right now, project permissions are at the same level that OUs or repos but it is possible to add them as subnodes of each repo.

Anyway, I'd wait for @mrietveld implementation to decide what the best strategy.

guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/ProjectSearchServiceImpl.java
dgutierr: It will be fixed with @mrietveld implementation.

csadilek:
Jenkins please retest this.

dgutierr:
retest

PR #321, opened 2016-06-02
Remove unused BuildConfig class

livthomas:
There was BuildConfig class which had to be passed to the test project REST call but was not used anywhere. Since the REST call failed when this parameter was not provided, I decided to remove it after the discussion with @mrietveld.


livthomas:
I will update the documentation (together with some other fixes) when @mrietveld finishes the migration to AsciiDoc.

mrietveld:
+1 if we somehow end up needing this functionality, I'd rather start from scratch than keep in code that's not needed anymore. 

@manstis Would you do the honors? (Merge)

PR #322, opened 2016-06-06
[6.4.x][RHBRMS-2447] fix asset-mgmt-project kjar to work OOTB

psiroky:
* before this change, WB logs WARNing that it can not
   find the asset-mgmt kjar, and then successfully installs
   it itself (so the WARN was pretty misleading)

 * the kjar is now built via maven-invoker-plugin and without
   any parent. The filtering avoids hardcoded any versions.

Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/216.

@manstis, @Salaboy, @nmirasch could one of you please review this?

PR #323, opened 2016-06-07
[RHBRMS-2447] fix asset-mgmt-project kjar to work OOTB

psiroky:
* before this change, WB logs WARNing that it can not
   find the asset-mgmt kjar, and then successfully installs
   it itself (so the WARN was pretty misleading)

 * the kjar is now built via maven-invoker-plugin and without
   any parent. The filtering avoids hardcoded any versions.

Forward port of #322 

Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/217


manstis:
+1 to merge (when green).

PR #326, opened 2016-06-20
GUVNOR-2565: NullPointerException when deleting repository using REST API

dgutierr:
Fix + test

@mswiderski Can you verify please?

### DIFF
--- a/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
+++ b/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
@@ -79,7 +79,7 @@ public class OrganizationalUnitServiceImpl implements OrganizationalUnitService 
     @Inject
     private Event<UpdatedOrganizationalUnitEvent> updatedOrganizationalUnitEvent;
 
-    private Map<String, OrganizationalUnit> registeredOrganizationalUnits = new HashMap<String, OrganizationalUnit>();
+    Map<String, OrganizationalUnit> registeredOrganizationalUnits = new HashMap<String, OrganizationalUnit>();
 
     @Inject
     private AuthorizationManager authorizationManager;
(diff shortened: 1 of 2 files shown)


manstis:
+1 to merge.

guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
```diff
@@ -114,6 +114,11 @@ public OrganizationalUnit getOrganizationalUnit( final String name ) {
 
     @Override
     public Collection<OrganizationalUnit> getAllOrganizationalUnits() {
+        return new ArrayList<>( registeredOrganizationalUnits.values() );
+    }
+
+    @Override
+    public Collection<OrganizationalUnit> getOrganizationalUnits() {
```
psiroky: Could we use more descriptive name? When looking at the code, it is not obvious to me what is the difference between `getAllOrganizationalUnits()` and `getOrganizationalUnits()`. Perhaps something like `getAccessibleOrganizationUnits`?

psiroky:
Verified that this fixes the currently failing REST tests in `kie-wb-distributions/kie-wb-tests/kie-wb-tests-rest`.

guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
dgutierr: I tried not to break API compatibility, that's why I kept the same method name. Anyway, provided javadoc is quite self-explanatory.

psiroky: Ah! I got confused by the diff, sorry. (now I also see the javadoc in the interface).

psiroky:
+1, merging as it is needed to fix currently failing REST tests.

PR #327, opened 2016-06-22
Fix test class location + remove unused deps + use logback-test.xml

psiroky:
* the test class was not in the standard location src/test/java
   (the 'java' dir was missing), thus the test was never executed

 * removed test-scoped deps are not needed, probably a result
   of a wrong copy&paste


manstis:
+1 to merge. Thanks @psiroky

PR #331, opened 2016-06-29
RHBPMS-3997: [GSS](6.3.x) Project dependencies are not retained in BPM Suite 6.3.0

karreiro:
This PR adds a method that retrieves compile dependencies. As "compile" is the default scope, this PR makes dependencies without scope be handled as a compile dependency.


Rikkola:
+1

PR #332, opened 2016-06-29
[6.4.x] RHBPMS-3997: [GSS](6.3.x) Project dependencies are not retained in BPM Suite 6.3.0

karreiro:
This PR adds a method that retrieves compile dependencies. As "compile" is the default scope, this PR makes dependencies without scope be handled as a compile dependency.

### DIFF
--- a/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/model/Dependencies.java
+++ b/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/model/Dependencies.java
@@ -75,6 +75,10 @@ public class Dependencies
         }
     }
 
+    public Collection<GAV> getCompileScopedGavs() {
+        return getGavs( "compile", null );
+    }
+
     @Override public int size() {
         return dependencies.size();
     }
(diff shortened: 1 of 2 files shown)


Rikkola:
+1

psiroky:
@karreiro could you please remove the `6.3.x` string from the commit message? Since this is for 6.4.x branch it could cause certain confusion (and it does not seem to add any additional info anyway).

(It also seems the `[GSS]` prefix is not necessary...)

Rikkola:
@psiroky I think gss and 6.3.x are there just because that is the Jira title.

psiroky:
@Rikkola sure, but I believe we don't need to copy the JIRA title into the commit message (at least not in cases where it seems wrong)

manstis:
@Rikkola @karreiro I'm merging this PR "as is".. but please consider @psiroky comment next time. I'm merging "as is" since (although misleading) it is consistent with https://github.com/droolsjbpm/kie-wb-common/pull/398 (which was not objected to and has been merged already - reverting and recommitting is going to be even more confusing).

PR #333, opened 2016-07-05
RHBRMS-2476: Allow users to edit project's GAV information in business central Project Editor when it has a parent

karreiro:
This PR adds a new system property that allows users to edit GAV information when it has a parent GAV.

### DIFF
--- a/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/service/ProjectRepositoryResolver.java
+++ b/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/service/ProjectRepositoryResolver.java
@@ -31,6 +31,8 @@ public interface ProjectRepositoryResolver {
 
     String CONFLICTING_GAV_CHECK_DISABLED = "org.guvnor.project.gav.check.disabled";
 
+    String CHILD_GAV_EDIT_ENABLED = "org.guvnor.project.gav.child.edit.enabled";
+
     /**
      * Get a collection of Repositories a Project will resolve artifacts against. The list will include
      * any Repositories defined in settings.xml
--- a/guvnor-project/guvnor-project-client/src/main/java/org/guvnor/common/services/project/client/POMEditorPanel.java
+++ b/guvnor-project/guvnor-project-client/src/main/java/org/guvnor/common/services/project/client/POMEditorPanel.java
@@ -22,6 +22,7 @@ import javax.inject.Inject;
 
 import com.google.gwt.user.client.ui.IsWidget;
 import com.google.gwt.user.client.ui.Widget;
+import org.guvnor.common.services.project.client.util.ApplicationPreferences;
 import org.guvnor.common.services.project.model.POM;
 import org.jboss.errai.ioc.client.container.SyncBeanDef;
 import org.jboss.errai.ioc.client.container.SyncBeanManager;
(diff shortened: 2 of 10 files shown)


guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/service/ProjectRepositoryResolver.java
```diff
@@ -31,6 +31,8 @@
 
     String CONFLICTING_GAV_CHECK_DISABLED = "org.guvnor.project.gav.check.disabled";
 
+    String CHILD_GAV_EDITION_ENABLED = "org.guvnor.project.gav.child.edit.enabled";
```
manstis: ```CHILD_GAV_EDIT_ENABLED``` would be more in keeping with the System Property and function... but it is me being a bit picky....

guvnor-project/guvnor-project-client/src/main/java/org/guvnor/common/services/project/client/POMEditorPanel.java
```diff
@@ -58,18 +59,21 @@ public void setPOM( POM model,
...
+        view.enableArtifactID();
+        view.enableVersion();
+
         if ( model.hasParent() ) {
             view.setParentGAV( model.getParent() );
             view.showParentGAV();
-            view.disableGroupID( "" );
-            view.enableArtifactID();
-            view.disableVersion( "" );
+            if ( !isChildGAVEditionEnabled() ) {
```
manstis: Your use of the word "edition" is wrong (but then again, don't even ask me what it is is Portuguese!) Personally I'd replace all occurrences of "edition" with "edit" e.g. ```isChildGAVEditEnabled()``` etc

guvnor-project/guvnor-project-client/src/main/java/org/guvnor/common/services/project/client/ProjectEntryPoint.java
```diff
@@ -0,0 +1,52 @@
...
+        this.appConfigService = appConfigService;
+    }
+
+    @AfterInitialization
+    public void startApp() {
+        loadPreferences();
+    }
+
+    private void loadPreferences() {
+        appConfigService.call( new RemoteCallback<Map<String, String>>() {
```
manstis: As an aside, you could use lamdas here instead..

manstis:
Nice test coverage.. just a minor comment about "edit" vs "edition"... would you mind changing please?

karreiro:
jenkins retest this

karreiro:
jenkins retest this

karreiro:
Hi @manstis,

I've already done the minor fix about "edit" vs "edition". However, when I pushed the commit, the build started becoming unstable.

Every tests are passing localy, but two of them are not passing at the CI. :(

For while I' removed the following assertions (that are working localy):

```
// guvnor-project/guvnor-project-client/src/test/java/org/guvnor/common/services/project/client/POMEditorPanelTest.java

@@ -102,6 +102,9 @@ public class POMEditorPanelTest {
         verify( view ).setGAV( gavModel.getGav() );
         verify( view ).setTitleText( "artifact" );
         verify( view ).setParentGAV( gavModel.getParent() );
-        verify( view ).disableGroupID( "" );
-        verify( view ).enableArtifactID();
-        verify( view ).disableVersion( "" );
         verify( view ).showParentGAV();
     }
 
@@ -162,6 +165,8 @@ public class POMEditorPanelTest {
         verify( view ).enableArtifactID();
         verify( view ).enableVersion();
         verify( view ).showParentGAV();
-        verify( view ).disableGroupID( anyString() );
-        verify( view ).disableVersion( anyString() );
     }
```

I think that the problem is something related to the way that I'm mocking and that would be a good idea to "resolve" the ticket to "unblock" it. And after that (in fact, right now) I can start investigating why this test has that strange behavior.

manstis:
@karreiro Hi.. it's probably caused by the ```ApplicationPreferences``` test that sets ```CHILD_GAV_EDIT_ENABLED``` to ```true```. Since ```ApplicationPreferences``` uses a ```static``` map the preferences are being retained across all tests. I'd suggest re-initialising the preferences to a "known state" in a ```@Before``` method..

manstis:
Yay, green.. +1 to merge.

PR #337, opened 2016-07-08
GUVNOR-2598: NPE when creating a Project in a Repository cloned via REST

manstis:
See https://issues.jboss.org/browse/GUVNOR-2598


csadilek:
+1

PR #338, opened 2016-07-13
[RHBRMS-2447] guvnor-asset-mgmt-project: replace also main POM to mak…

psiroky:
…e the build consistent

 * leaving the original POM caused issues with BREW since it checks
   for the consistency between the deployed POM and the one bundled
   inside the jar

Depends on: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/236

### DIFF
--- a/guvnor-asset-mgmt/guvnor-asset-mgmt-project/pom.xml
+++ b/guvnor-asset-mgmt/guvnor-asset-mgmt-project/pom.xml
@@ -10,7 +10,6 @@
   </parent>
 
   <artifactId>guvnor-asset-mgmt-project</artifactId>
-  <packaging>pom</packaging>
 
   <name>Guvnor - Asset Management Project</name>
   <description>Guvnor - Asset Management Project (includes processes to manage business assets)</description>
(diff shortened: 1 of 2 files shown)


manstis:
@psiroky Hi. who's best to review this? I know the issue that lead to the original (and subsequent) PRs.. but I don't really understand the revised Maven configuration...

mswiderski:
fyi - just a side note as it only concerns master, assetmgmt is going away soon as part of remove of jbpm runtime from workbench - in case this change is also planned for master

psiroky:
@manstis not sure who would be the best. I am not really proud of myself for coming up with this hack, but I could not find any better solution. I agree the Maven setup is far from easy to follow. I can try to explain it to you (and then also amend the comment in the POM so others reading this in future would understand as well).

@mswiderski if by soon you day or two, I will not cherry-pick this into master. Otherwise I think it would be better to have the branches synced to avoid any possible confusion.

mswiderski:
we are doing final tests after squashing commits so I do believe it should be one or two days till PRs are sent but then it's review time...

manstis:
+1 to merge (when green).

PR #339, opened 2016-07-13
[RHBRMS-2490] guvnor-asset-mgmt-project: replace also main POM to mak…

psiroky:
…e the build consistent

 * leaving the original POM caused issues with BREW since it checks
   for the consistency between the deployed POM and the one bundled
   inside the jar

Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/237


manstis:
@psiroky Likewise (with the 6.4.x PR).. if you want to talk me through the change I'll happily +1...

manstis:
+1 to merge (when green).

PR #340, opened 2016-07-13
UF-333: Stored values are not correctly presented 

dgutierr:
Fix + tests

@manstis Can you review please?


manstis:
+1 to merge (when green)

PR #343, opened 2016-07-14
JBPM-5071 - remove of jbpm runtime from workbench

mswiderski:
- Migrate executor service from Kie to Uberfire
- clean up of asset mgmt - leaving only create repository with branches


mrietveld:
+1 for the guvnor-rest changes, which look logical. 

However, I don't know the UberFire executor code that well, so the real test will be the integration tests.

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/AssetManagementServiceImpl.java
```diff
@@ -15,73 +15,72 @@
...
-        this.promoteChangesEvent = promoteChangesEvent;
-        this.releaseProjectEvent = releaseProjectEvent;
-        this.configurationService = configurationService;
+        this.pomService = pomService;
+        this.repositoryService = repositoryService;
         this.projectService = projectService;
     }
 
     @PostConstruct
     public void init() {
```
manstis: Do we need an empty ```@PostConstruct``` method?

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/AssetManagementServiceImpl.java
```diff
@@ -91,111 +90,87 @@ public void configureRepository( final String repository,
...
     public Set<Project> getProjects( final Repository repository,
                                      final String branch ) {
         return projectService.get().getProjects( repository,
                                                  branch );
     }
 
-    protected String encodePassword( final String password ) {
-        if ( password == null ) {
-            return null;
+    private Set<Project> getProjects( final Repository repository ) {
```
manstis: The old [code](https://github.com/mswiderski/guvnor/blob/44651ca6a7fa8b147212302b4f47e8408b91c7ad/guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/AssetManagementServiceImpl.java#L190) would get a ```Set``` of "[authorized](https://github.com/droolsjbpm/guvnor/blob/master/guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/AbstractProjectService.java#L145)" projects; this new code simply gets a collection of all Projects in the Repository.

manstis:
A couple of comments.. otherwise looks good to me.

Thanks @mrietveld for looking at the REST related changes too,

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/AssetManagementServiceImpl.java
mswiderski: that's actually as it should be as this is system operation so all projects should be updated to be consistent

guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/AssetManagementServiceImpl.java
mswiderski: removed

manstis:
Thanks @mswiderski +1 to merge (please let me know when this can be merged).

PR #360, opened 2016-09-09
[6.5.x] moved setAssetsManagementGrant check to showForm to make sure require…

mswiderski:
…d roles are loaded in KieACL

Recently it was few times reported that there is no option to mark repo as managed while cloning based on the investigation it turned out that the check was done to early - before the roles are loaded from workbench policy and thus resulting always as disabled.

Tests for clone repo are already there but not sure this position change should be tested or not...

### DIFF
--- a/guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
+++ b/guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/clone/CloneRepositoryPresenter.java
@@ -88,7 +88,6 @@ public class CloneRepositoryPresenter implements CloneRepositoryView.Presenter {
     public void init() {
         view.init( this,
                    isOuMandatory() );
-        setAssetsManagementGrant();
     }
 
     @AfterInitialization
@@ -238,6 +237,7 @@ public class CloneRepositoryPresenter implements CloneRepositoryView.Presenter {
 
     public void showForm() {
         view.reset();
+        setAssetsManagementGrant();
         view.show();
     }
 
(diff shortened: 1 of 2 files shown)


manstis:
@mswiderski Nice catch. 

A test to verify ```setAssetsManagementGrant()``` is called by ```showForm()``` wouldn't go amiss.

mswiderski:
@manstis would this be what you had in mind: https://github.com/droolsjbpm/guvnor/pull/360/files#diff-1c60ac86dc6308928a900010a1adb84eR395

manstis:
@mswiderski Yep.. it'll prevent future sabotage! 

Interestingly this fix only really makes the "enable/disable" happen later; but still no *guarantee* it will happen *after* ```KieACL``` is initialised. I wonder if @dgutierr has introduced a "startup blocker" on ```KieACL``` to prevent the workbench from running until ```KieACL``` is setup?

@dgutierr We have a similar "startup blocker" for "Social Activity" bootstrap [here](https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/client/KieWorkbenchEntryPoint.java#L126). Do we have any existing guarantee on master that ```KieACL``` would have been initialised before the workbench runs?

manstis:
Anyway.. for the purpose of this PR.. 

+1 to merge (when green).. we've *minimised* the likelihood of the issue.

dgutierr:
@manstis I'm afraid we are not using a startup blocker in 6.5 for the initialization on the ACL.

https://github.com/droolsjbpm/kie-wb-distributions/blob/6.5.x/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/client/KieWorkbenchEntryPoint.java#L139

mswiderski:
seems like there is some build server issue but... jenkins retest this

mswiderski:
this would be good to include in upcoming release so it would be tested by community as well

PR #361, opened 2016-09-10
[UF-403] Fork and Mirror

adrielparedes:
The goal is to have the possibility to fork any git repository inside the niogit directory.
To complete the goal we need:
* Separate git repositories into different directories. Those directories are named like the Organizational Units.
* Create a Fork command that copy all the content of the repository to fork and add the reference to the internal path of the origin repository.
* Write information about fork so we can track where those forks are, how many forks has those repositories, etc.

### DIFF
--- a/guvnor-structure/guvnor-structure-api/src/main/java/org/guvnor/structure/repositories/impl/metadata/GitMetadataImpl.java
+++ b/guvnor-structure/guvnor-structure-api/src/main/java/org/guvnor/structure/repositories/impl/metadata/GitMetadataImpl.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.structure.repositories.impl.metadata;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.guvnor.structure.repositories.GitMetadata;
+import org.jboss.errai.common.client.api.annotations.MapsTo;
+import org.jboss.errai.common.client.api.annotations.Portable;
+
+import static org.uberfire.commons.validation.PortablePreconditions.*;
+
+@Portable
+public class GitMetadataImpl implements GitMetadata {
+
+    private String origin;
+    private List<String> forks;
+    private String name;
+
+    public GitMetadataImpl( String name ) {
+        this( name, "", new ArrayList<>() );
+    }
+
+    public GitMetadataImpl( String name,
+                            String origin ) {
+        this( name, origin, new ArrayList<>() );
+    }
+
+    public GitMetadataImpl( String name,
+                            List<String> forks ) {
+        this( name, "", forks );
+    }
+
+    public GitMetadataImpl( @MapsTo("name") String name,
+                            @MapsTo("origin") String origin,
+                            @MapsTo("forks") List<String> forks ) {
+        this.name = checkNotEmpty( "name", name );
+        this.origin = checkNotNull( "origin", origin );
+        this.forks = checkNotNull( "forks", forks );
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public List<String> getForks() {
+        return new ArrayList<>( forks );
+    }
+
+    @Override
+    public String getOrigin() {
+        return origin;
+    }
+
+}
--- a/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/RepositoryServiceImpl.java
+++ b/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/RepositoryServiceImpl.java
@@ -29,6 +29,7 @@ import org.guvnor.common.services.backend.exceptions.ExceptionUtilities;
 import org.guvnor.structure.backend.backcompat.BackwardCompatibleUtil;
 import org.guvnor.structure.organizationalunit.OrganizationalUnit;
 import org.guvnor.structure.organizationalunit.OrganizationalUnitService;
+import org.guvnor.structure.repositories.GitMetadataStore;
 import org.guvnor.structure.repositories.NewRepositoryEvent;
 import org.guvnor.structure.repositories.Repository;
 import org.guvnor.structure.repositories.RepositoryAlreadyExistsException;
@@ -70,6 +71,9 @@ public class RepositoryServiceImpl implements RepositoryService {
     @Inject
     @Named("ioStrategy")
     private IOService ioService;
+
+    @Inject
+    private GitMetadataStore metadataStore;
 
     @Inject
     private ConfigurationService configurationService;
@@ -200,9 +204,9 @@ public class RepositoryServiceImpl implements RepositoryService {
     @Override
     public Collection<Repository> getRepositories() {
         Collection<Repository> result = new ArrayList<>();
-        for (Repository repository : configuredRepositories.getAllConfiguredRepositories()) {
-            if (authorizationManager.authorize(repository, sessionInfo.getIdentity())) {
-                result.add(repository);
+        for ( Repository repository : configuredRepositories.getAllConfiguredRepositories() ) {
+            if ( authorizationManager.authorize( repository, sessionInfo.getIdentity() ) ) {
+                result.add( repository );
             }
         }
         return result;
@@ -215,12 +219,14 @@ public class RepositoryServiceImpl implements RepositoryService {
                                         final RepositoryEnvironmentConfigurations repositoryEnvironmentConfigurations ) throws RepositoryAlreadyExistsException {
 
         try {
+
             final Repository repository = createRepository( scheme,
                                                             alias,
                                                             repositoryEnvironmentConfigurations );
             if ( organizationalUnit != null && repository != null ) {
                 organizationalUnitService.addRepository( organizationalUnit, repository );
             }
+            metadataStore.write( alias, (String) repositoryEnvironmentConfigurations.getOrigin() );
             return repository;
         } catch ( final Exception e ) {
             logger.error( "Error during create repository", e );
@@ -263,6 +269,7 @@ public class RepositoryServiceImpl implements RepositoryService {
                     if ( repository.getAlias().equals( alias ) ) {
                         organizationalUnitService.removeRepository( ou,
                                                                     repository );
+                        metadataStore.delete( alias );
                     }
                 }
             }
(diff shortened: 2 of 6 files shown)


manstis:
ok to test

guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/RepositoryServiceImpl.java
```diff
@@ -263,6 +281,8 @@ public void removeRepository( final String alias ) {
                     if ( repository.getAlias().equals( alias ) ) {
                         organizationalUnitService.removeRepository( ou,
                                                                     repository );
+                        String metadata = ou.getName() + "/" + alias;
```
manstis: Won't this fail if the ```finalAlias``` was originally created with [this](https://github.com/droolsjbpm/guvnor/pull/361/files#diff-0d90a1d4d7ccbb7bb35bb56aad51f4feR245)?

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryPresenter.java
```diff
@@ -0,0 +1,203 @@
...
+        this.placeManager = placeManager;
+        this.authorizationManager = authorizationManager;
+        this.sessionInfo = sessionInfo;
+    }
+
+    @PostConstruct
+    public void init() {
+        view.init( this,
+                   isOuMandatory() );
+        setAssetsManagementGrant();
```
manstis: Please see [here](https://github.com/droolsjbpm/guvnor/commit/083845d8bdda17ff002544346af029c21ec6e386). The call to ```setAssetsManagementGrant()``` was moved to (in your case) ```showForm()``` to ensure ```KieACL``` has had time to load and initialise.

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryPresenter.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    private RemoteCallback<String> getNormalizeRepositoryNameCallback() {
+        return new RemoteCallback<String>() {
+            @Override
+            public void callback( final String normalizedName ) {
+                lockScreen();
+
+                final String scheme = "git";
+                final String[] source = view.getSourceSelectedOrganizationalUnit().split( "/" );
```
manstis: I did not see any validation of either "source Organisational Unit" nor" target Organisational Unit". Since both are ```TextBox``` in the UI; is it not possible that either could *not* contain a "/" and hence the split result in arrays of size one leading to the below use of index 1 to fail?

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryViewImpl.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    @Override
+    public void closeBusyPopup() {
+        BusyPopup.close();
+    }
+
+    @Override
+    public boolean showAgreeNormalizeNameWindow( final String normalizedName ) {
+        return Window.confirm( CoreConstants.INSTANCE.RepositoryNameInvalid() + " \"" + normalizedName + "\". " + CoreConstants.INSTANCE.DoYouAgree() );
```
manstis: For consistency would it not be better to use ```BaseModal``` and ```ModalFooterYesNoCancelButtons``` instead of ```Window.confirm```?

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryViewImpl.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    @Override
+    public void alertRepositoryForked() {
+        Window.alert( CoreConstants.INSTANCE.RepoCloneSuccess() + "\n\n" + CommonConstants.INSTANCE.IndexClonedRepositoryWarning() );
```
manstis: We should try to avoid use of ```Window.alert(...)``` and ```Window.confirm(...)``` as it's difficult to write Selenium tests for them - since different browsers implement these methods differently. Use of ```BaseModal``` and ```ModalFooterOKButton``` would be preferred (I realise some of this PR is copy and paste of old code.. but now is probably the best time to refactor).

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryPresenter.java
```diff
@@ -0,0 +1,203 @@
...
+import org.jboss.errai.common.client.api.RemoteCallback;
+import org.uberfire.client.mvp.PlaceManager;
+import org.uberfire.mvp.impl.DefaultPlaceRequest;
+import org.uberfire.rpc.SessionInfo;
+import org.uberfire.security.authz.AuthorizationManager;
+
+import static org.guvnor.structure.security.RepositoryFeatures.*;
+
+@Dependent
+public class ForkRepositoryPresenter implements ForkRepositoryView.Presenter {
```
manstis: Please see ```org.guvnor.structure.client.editors.repository.clone.CloneRepositoryPresenter``` Organisational Units are previously been supplied as a "select list" for Users whereas this PR seems to make entry of OUs based on ```TextBox```. Is this deliberate? Can we not present Users with a list of OUs?

Furthermore the referenced class observes both ```AfterCreateOrganizationalUnitEvent``` and ```AfterDeleteOrganizationalUnitEvent``` to ensure such list is synchronised with changes made in the "OU Management" screen - also on the Administration Perspective (we've had JIRAs in the past for when the "Clone/New Repository" popups did not reflect the "current" state of available OUs).

manstis: There also appears to be a distinct lack of tests for this Presenter.

manstis:
@adrielparedes @porcelli Could one of you please give some background information about this PR? I assume it's preliminary work for "workspaces" and the ```ForkRepositoryPresenter``` temporary(?) to assist development/use of the upcoming feature?

csadilek:
@adrielparedes @porcelli @manstis subscribe. ;)

adrielparedes:
@adrielparedes @porcelli @manstis @csadilek Yes, sorry it isn't so clear. The ForkPresenter/View is based con ClonePresenter, and it's just to have a reference. It was made to test and see the real result of the Fork command. I understand that this view is not going to be part of the final platform.

manstis:
@adrielparedes Thanks for the explanation. So the guts of feature are in [Uberfire](https://github.com/uberfire/uberfire/pull/504) - that I'll let @porcelli review.

Unfortunately I have an issue with accepting this PR on master as it is not feature complete and in essence only a "debug" tool for further development of the workspace concept. 

If we must have this I'd prefer it to go into [drools-wb-webapp](https://github.com/droolsjbpm/drools-wb/tree/master/drools-wb-webapp/src/main/java/org/drools/workbench/client) which is *not* considered part of any distribution. You have a [PR](https://github.com/droolsjbpm/drools-wb/pull/288) for drools-wb anyway; so you'd be effectively moving the whole "debug tooling" into somewhere more transient.

adrielparedes:
@manstis it's ok for me, I'm going to move the presenter/view to the drools-wb-webapp, and I will write a comment. I will tell you when it's finished.

Thanks!

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryViewImpl.java
```diff
@@ -0,0 +1,232 @@
...
+import org.gwtbootstrap3.client.ui.FormLabel;
+import org.gwtbootstrap3.client.ui.TextBox;
+import org.gwtbootstrap3.client.ui.constants.ValidationState;
+import org.uberfire.ext.widgets.common.client.common.BusyPopup;
+import org.uberfire.ext.widgets.common.client.common.popups.BaseModal;
+import org.uberfire.ext.widgets.common.client.common.popups.errors.ErrorPopup;
+import org.uberfire.ext.widgets.core.client.resources.i18n.CoreConstants;
+
+@Dependent
+public class ForkRepositoryViewImpl extends BaseModal implements ForkRepositoryView,
```
porcelli: -1 for UI at this point, but if you want add UI, it has to be based on Errai UI

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryViewImpl.ui.xml
```diff
@@ -0,0 +1,77 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
+<ui:UiBinder xmlns:ui="urn:ui:com.google.gwt.uibinder"
```
porcelli: -1 for UI at this point, but if you want add UI, it has to be based on Errai UI

guvnor-structure/guvnor-structure-client/src/main/java/org/guvnor/structure/client/editors/repository/fork/ForkRepositoryPresenter.java
porcelli: -1 for UI at this point, but if added has to take @manstis points in considerations.

adrielparedes:
@porcelli if this View is not going to be used, I think that I prefer to delete them so I can concentrate in fork and track feature until I get better with Errai, what do you think?

porcelli:
@adrielparedes up to you. I've been always thinking this as a non-ui work for now. this is more plumbing infra.. that we'll build on top.

guvnor-structure/guvnor-structure-api/src/main/java/org/guvnor/structure/repositories/impl/metadata/GitMetadataImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    private String name;
+
+    public GitMetadataImpl() {
+    }
+
+    public GitMetadataImpl( String name,
+                            String origin ) {
+        this.name = name;
+        this.origin = origin;
+    }
```
porcelli: here forks are null; better initialize it with a new empty ArrayList<String>

adrielparedes: @porcelli should I check for the others parameters also?

porcelli: i'd add a checkNotEmpty() where can't be empty

guvnor-structure/guvnor-structure-api/src/main/java/org/guvnor/structure/repositories/impl/metadata/GitMetadataImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public List<String> getForks() {
+        if ( this.forks == null ) {
```
porcelli: this can't happen

guvnor-structure/guvnor-structure-api/src/main/java/org/guvnor/structure/repositories/impl/metadata/GitMetadataImpl.java
```diff
@@ -0,0 +1,76 @@
...
+            this.forks = new ArrayList<>();
+        }
+        return new ArrayList<>( forks );
+    }
+
+    @Override
+    public String getOrigin() {
+        if ( this.origin == null ) {
```
porcelli: can't be null

PR #362, opened 2016-09-14
GUVNOR-2636: [Decision Table Spreadsheet] uploading new version causes error

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2636

### DIFF
--- a/guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/file/upload/AbstractFileServlet.java
+++ b/guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/file/upload/AbstractFileServlet.java
@@ -24,9 +24,11 @@ import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Iterator;
 import java.util.List;
+import javax.servlet.ServletConfig;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
 
 import org.apache.commons.fileupload.FileItem;
 import org.apache.commons.fileupload.FileItemFactory;
@@ -35,11 +37,16 @@ import org.apache.commons.fileupload.servlet.ServletFileUpload;
 import org.apache.commons.io.IOUtils;
 import org.guvnor.common.services.shared.file.upload.FileManagerFields;
 import org.guvnor.common.services.shared.file.upload.FileOperation;
+import org.jboss.errai.bus.client.api.QueueSession;
+import org.jboss.errai.bus.server.api.SessionProvider;
+import org.jboss.errai.bus.server.servlet.ServletBootstrapUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.uberfire.backend.server.util.Paths;
 import org.uberfire.backend.vfs.Path;
 import org.uberfire.server.BaseFilteredServlet;
+
+import static org.uberfire.commons.validation.PortablePreconditions.*;
 
 /**
  * This is for dealing with assets that have an attachment (ie assets that are really an attachment).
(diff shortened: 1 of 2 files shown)


guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/file/upload/AbstractFileServlet.java
```diff
@@ -294,4 +297,12 @@ protected void processAttachmentDownload( final String url,
         }
     }
 
+    protected String getSessionId( final HttpServletRequest request,
+                                   final SessionProvider sessionProvider ) {
+        final HttpSession session = request.getSession( true );
+        final String clientId = request.getParameter( "clientId" );
```
manstis: I think there is (at least) another "file upload" screen (in ```kie-wb-common```) that you have not (yet) fixed to include a ```clientId```. Therefore there is potential for this to be ```null``` which may cause problems with ```SessionProvider``` (probably will!). I'll try to find the screen - but for now please check for *all* usage of ```AttachmentFileWidget```

manstis:
I'd be advisable to protect against a ```null``` ```clientId``` (in case we miss a "upload" screen to fix) otherwise looks good to me..

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/file/upload/AbstractFileServlet.java
manstis: @karreiro There's ```org.kie.workbench.common.screens.defaulteditor.client.editor.NewFileUploader``` that'll probably need looking at too :-(

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/file/upload/AbstractFileServlet.java
```diff
@@ -294,4 +309,12 @@ protected void processAttachmentDownload( final String url,
         }
     }
 
+    protected String getSessionId( final HttpServletRequest request,
+                                   final SessionProvider sessionProvider ) {
+        final HttpSession session = request.getSession( true );
+        final String clientId = checkNotNull( "clientId", request.getParameter( "clientId" ) );
```
manstis: Personally I'd prefer a fall-back if ```request.getParamete("clientId")``` returned ```null```. For example default to zero and log a warning that the parameter was not found. The worst it can do is mean the concurrent update mechanism will not work properly. Better than throwing an ```IllegalArgumentException``` IMO.

manstis:
Jenkins please retest this.

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/file/upload/AbstractFileServlet.java
karreiro: That sounds like better a approach. Thanks @manstis

guvnor-services/guvnor-services-backend/src/test/java/org/guvnor/common/services/backend/file/upload/AbstractFileServletTest.java
```diff
@@ -0,0 +1,99 @@
...
+    private HttpServletRequest request;
+
+    @Mock
+    private QueueSession queueSession;
+
+    @Mock
+    private HttpSession httpSession;
+
+    @Test
+    public void testGetSessionId() throws Exception {
```
manstis: @karreiro You know what I'm going to ask, don't you?..... How about a test for when the request parameter is missing?

karreiro: Thanks @manstis

manstis:
Sweet @karreiro It's a +1 from me.. (when green).

PR #364, opened 2016-09-20
[UF-403]: Fix added to solve origin null parameter problem

adrielparedes:
@manstis this solves the problem for: _java.lang.IllegalArgumentException: Parameter named 'origin' should be not null!_

I've added a test also, to check that behavior.


porcelli:
+1

PR #365, opened 2016-09-20
[6.5.x] RHBPMS-4275: Guided Decision Table: XLS can not be validated

manstis:
See https://issues.jboss.org/browse/RHBPMS-4275

### DIFF
--- a/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/builder/service/BuildService.java
+++ b/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/builder/service/BuildService.java
@@ -37,16 +37,6 @@ public interface BuildService {
     BuildResults build( final Project project );
 
     /**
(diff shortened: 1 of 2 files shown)


manstis:
Part of an ensemble:

- https://github.com/droolsjbpm/guvnor/pull/365
- https://github.com/droolsjbpm/kie-wb-common/pull/476
- https://github.com/droolsjbpm/drools-wb/pull/298

manstis:
@karreiro Hi, I had to make a change to the revised "Asset Validation" code since XLS Decision Tables cannot be converted to ```String``` for validation (as we try to convert the XLS file to text.. which, unsurprisingly, doesn't give the correct results). I toyed with converting the XLS to DRL and pass this to the generic validation; however we'd need to hack the path to trick KIE into thinking it was ```.drl``` and not  ```.xls``` (otherwise KIE passes the content through a XLS->DRL converter). 

Could you please review (all of your existing tests continue to pass).

karreiro:
+1

PR #379, opened 2016-10-24
UF-320: Improvements on admin page and preferences view

paulovmr:
* Mapped workbench specific preference scopes.

### DIFF
--- a/guvnor-services/guvnor-services-api/src/main/java/org/guvnor/common/services/shared/preferences/WorkbenchPreferenceScopeResolutionStrategies.java
+++ b/guvnor-services/guvnor-services-api/src/main/java/org/guvnor/common/services/shared/preferences/WorkbenchPreferenceScopeResolutionStrategies.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.common.services.shared.preferences;
+
+import org.uberfire.ext.preferences.shared.impl.PreferenceScopeResolutionStrategyInfo;
+
+/**
+ * Service used to obtain information regarding the workbench preference scopes resolution strategies.
+ */
+public interface WorkbenchPreferenceScopeResolutionStrategies {
+
+    /**
+     * Returns a scope resolution strategy with the scopes (in that order): scope (parameter), user and global.
+     * @param scope The first scope in the resolution order (e.g. project).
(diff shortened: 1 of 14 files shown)


paulovmr:
Part of an ensemble:

* https://github.com/uberfire/uberfire/pull/538
* https://github.com/droolsjbpm/guvnor/pull/379
* https://github.com/droolsjbpm/kie-wb-common/pull/519
* https://github.com/droolsjbpm/kie-wb-distributions/pull/381

guvnor-services/guvnor-services-api/src/main/java/org/guvnor/common/services/shared/preferences/WorkbenchPreferenceScopeResolutionStrategies.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.common.services.shared.preferences;
+
+import org.uberfire.ext.preferences.shared.impl.PreferenceScopeResolutionStrategyInfo;
+
```
csadilek: javadoc, please ;)

csadilek:
@paulovmr shouldn't this all be in kie-wb-common instead of guvnor?

paulovmr:
@csadilek Thanks for the review. I left this in guvnor because we will need some preferences in guvnor, and kie-wb-common depends on guvnor, not the other way.

csadilek:
@paulovmr this PR sets up the default preference scope resolutions strategy for the WBs, right? I really think should be in kie-wb-common. 

Might be missing something, but where are the preferences in this PR used by guvnor? Couldn't those be added from kie-wb-common as well (as it depends on Guvnor). Alternatively, the WB could add those preferences tools/screens directly (as it also depends on Guvnor), no ? It just seems that this would be a cleaner separation of concerns.

paulovmr:
@csadilek No preference is created in this PR, indeed. 

Before start doing this, I had the gut feeling of putting this on kie-wb-common as well, but then I thought that we could have to migrate some guvnor preferences (e.g. [ApplicationPreferences](https://github.com/droolsjbpm/guvnor/blob/master/guvnor-project/guvnor-project-client/src/main/java/org/guvnor/common/services/project/client/util/ApplicationPreferences.java) to the new API, and if the scope resolution strategy was in kie-wb-common, we wouldn't be able to read their values from guvnor classes. But maybe there is a way to do it that I didn't think of...

csadilek:
@paulovmr understood, but these guvnor preferences only ever need to be retrieved from within the WB, right? The resolution strategy will then be set up anyway (by kie-wb-common).

paulovmr:
@csadilek I see your point, yes you are right. But, in the near future, I'm thinking about adding in Uberfire a way to pass a custom scope resolution strategy through the beans API, what means that the [BasePreference](https://github.com/uberfire/uberfire/blob/master/uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/BasePreference.java) would have `load` and `save` methods that receive a custom scope resolution strategy as a parameter. This would be necessary if we want to retrieve preferences of specific projects in the code. 

With that in mind, we would have to access the classes in this PR on the guvnor code. Sorry, I should have thought about mentioning that before.

paulovmr:
@csadilek So, as we talked on IRC, I will close this PR and open another one on kie-wb-common.

PR #380, opened 2016-10-28
- Supporting Dev And Super Dev Mode for LS

Salaboy:
This pull request add the pipeline modification for supporting Dev and Super Dev Mode in LS. 
This PR includes the updated Repository visitor to work with temporary directories that are indexed to be reused and only the changes are applied when needed.
This PR is linked to this Drools PR that needs to be merged in in order for this code to work: https://github.com/droolsjbpm/drools/pull/968

and to this PR in LS: https://github.com/droolsjbpm/livespark/pull/57

@mbarkley can you please test this  and give feedback?


guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/MavenCliOutputTest.java
```diff
@@ -0,0 +1,148 @@
...
+        tempPath = Files.createTempDirectory("zzz").toFile();
+    }
+
+    @After
+    public void tearDown() {
+        FileUtils.deleteQuietly(tempPath);
+    }
+
+    @Test
+    public void waitForAppBuildTest() {
```
mbarkley: Am I correct in assuming that this only tests the `MavenBuilderExecutor.executeMaven` method? If that is the case I think a unit test with a mocked Maven executor would be easier to understand and maintain.

guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/RepositoryVisitorTest.java
```diff
@@ -0,0 +1,139 @@
...
+        MapDifference<String, String> difference = Maps.difference(identityHash, newIdentityHash);
+
+        Map<String, MapDifference.ValueDifference<String>> entriesDiffering = difference.entriesDiffering();
+        System.out.println(" Size of Differences: " + entriesDiffering.size());
+        for (String key : entriesDiffering.keySet()) {
+            System.out.println("Different Value: " + key);
+        }
+        assertEquals(1, entriesDiffering.size());
+
+        Map<String, String> deletedFiles = difference.entriesOnlyOnLeft();
```
mbarkley: I think you need to assert how many, and what files were deleted.

guvnor-ala/guvnor-ala-distribution/pom.xml
```diff
@@ -16,452 +16,433 @@
...
+    <version>7.0.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>guvnor-ala-distribution</artifactId>
+  <packaging>war</packaging>
+  <name>Guvnor ALA :: Distribution (Swarm/Docker)</name>
+  <properties>
+    <endorsed.dir>${project.build.directory}/endorsed</endorsed.dir>
+  </properties>
+  <dependencies>
+    <!--&lt;!&ndash; Java EE 7 dependency &ndash;&gt;-->
```
mbarkley: Why is this commented out? Should it be here or should it be removed?

guvnor-ala/guvnor-ala-spi/src/main/java/org/guvnor/ala/registry/SourceRegistry.java
```diff
@@ -34,8 +33,8 @@
      * @param Repository meta data
      * @see Repository
      */
-    void registerRepositorySources( final Path path,
-            final Repository repo );
+    void registerRepositorySources(final Path path,
```
mbarkley: I think you've used the wrong formatting for methods in this file (they should have spaces around arguments, right?).

guvnor-ala/guvnor-ala-wildfly-provider/src/test/java/org/guvnor/ala/wildfly/executor/tests/WildflyExecutorTest.java
```diff
@@ -89,129 +89,224 @@
     @Before
     public void setUp() {
         try {
-            tempPath = Files.createTempDirectory( "xxx" ).toFile();
-        } catch ( IOException e ) {
+            tempPath = Files.createTempDirectory("xxx").toFile();
+        } catch (IOException e) {
```
mbarkley: More incorrect formatting in this file.

mbarkley:
Jenkins, retest this.

guvnor-ala/guvnor-ala-wildfly-provider/src/test/java/org/guvnor/ala/wildfly/executor/tests/WildflyExecutorTest.java
```diff
@@ -89,129 +89,224 @@
...
             e.printStackTrace();
         }
     }
 
     @After
```
porcelli: @AfterClass! 
otherwise you're deleting the repo, and UF will get lost.

guvnor-ala/guvnor-ala-wildfly-provider/src/test/java/org/guvnor/ala/wildfly/executor/tests/WildflyExecutorTest.java
```diff
@@ -89,129 +89,224 @@
     @Before
```
porcelli: @BeforeClass!

guvnor-ala/guvnor-ala-wildfly-provider/src/test/java/org/guvnor/ala/wildfly/executor/tests/WildflyRuntimeTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @InSequence(2)
+    public void waitForAppBuildTest() {
+        final GitHub gitHub = new GitHub();
+        final GitRepository repository = (GitRepository) gitHub.getRepository( "salaboy/drools-workshop", new HashMap<String, String>() {
+            {
+                put( "out-dir", tempPath.getAbsolutePath() );
+            }
+        } );
+        final Source source = repository.getSource( "master" );
+        assertNotNull( source );
```
porcelli: All the above is not the most correct way to do what you want, this is more what you're looking for:

```java
        final Optional<Source> _source = new GitConfigExecutor( new InMemorySourceRegistry() ).apply( new GitConfigImpl( tempPath.getAbsolutePath(),
                                                                                                                  "master",
                                                                                                                  "https://github.com/salaboy/drools-workshop",
                                                                                                                  "drools-workshop",
                                                                                                                  "true" ) );

        assertTrue( _source.isPresent() );
        final Source source = _source.get();

```

guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/MavenCliOutputTest.java
Salaboy: @mbarkley, this test is also testing that the output for the build and that the streams are working as expected. I would leave the test as it is if possible. Do you think that we can add extra assertions to make it more clear?

guvnor-ala/guvnor-ala-build-maven/src/main/java/org/guvnor/ala/build/maven/util/RepositoryVisitor.java
```diff
@@ -13,109 +13,221 @@
...
     }
 
-    private void makeTempRootDirectory() {
-        File tempDirectory = new File( buildRoot );
-        tempDirectory.mkdirs();
+    private File makeTempRootDirectory(final String tempRoot,
+            final boolean recreateTempDir) {
+        final File tempRootDir = new File(tempRoot);
+        if (tempRootDir.exists()) {
+            if (recreateTempDir) {
```
Salaboy: @mbarkley the recreateDir flag, just means to make sure that we don't reuse an existing dir, for that reason it just delete the directory, and the visitor will be in charge of repopulate all the files. I coudn't think of a better name based on what is doing.. mostly because I wanted to keep the name of the variable positive.. but I agree that is somewhat confusing.

mbarkley: Okay. So really this method is about cleaning a reused directory, right? If that's the case I would suggest renaming the method and parameter to use the word "clean" instead of "make" and "recreate".

guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/MavenCliOutputTest.java
Salaboy: +1.. I will do that

guvnor-ala/guvnor-ala-build-maven/src/main/java/org/guvnor/ala/build/maven/util/RepositoryVisitor.java
Salaboy: +1 thanks for suggesting the terms.

Salaboy:
@mbarkley PR updated.. can you take a look at it now?

mbarkley:
+1

PR #386, opened 2016-11-11
UF-320: Including workbench and project preferences

paulovmr:
* Extracting GAV preferences to the new preferences API.
* Project preferences: specific preferences for each project.


paulovmr:
Part of an ensemble:

* https://github.com/uberfire/uberfire/pull/558
* https://github.com/droolsjbpm/guvnor/pull/386
* https://github.com/droolsjbpm/kie-wb-common/pull/558
* https://github.com/droolsjbpm/kie-wb-distributions/pull/404

paulovmr:
I'm going to make some changes on this PR, please wait for review/merge.

guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/preferences/GAVPreferences.java
```diff
@@ -0,0 +1,67 @@
...
+            formType = PropertyFormType.BOOLEAN)
+    private boolean conflictingGAVCheckDisabled;
+
+    @Property(bundleKey = "GAVPreferences.ChildGAVEditEnabled.Label",
+            formType = PropertyFormType.BOOLEAN)
+    private boolean childGAVEditEnabled;
+
+    @Override
+    public GAVPreferences defaultValue( final GAVPreferences defaultValue ) {
+        final String conflictingGAVCheckDisabledSystemProperty = System.getProperty( "org.guvnor.project.gav.check.disabled", "false" );
```
manstis: This class is in "-api" and covered by the GWT module's ```<source>``` paths; so will be available client-side as well as server-side. Therefore the ```System.getProperty(...)``` calls are likely to fail on the client. The old way we made System Properties available client-side was with ```GAVPreferencesLoader``` which ran on the server and ultimately lead to the initialisation of  ```ApplicationPreferences``` on the client (see [here](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/entrypoint/DefaultWorkbenchEntryPoint.java#L82)). Can you please advise how the new mechanism works?

paulovmr: The only place where the `defaultValue` method is called (indirectly) is [this](https://github.com/uberfire/uberfire/blob/master/uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/PreferenceBeanDefaultValueRecorder.java#L45), to set the global value of the preference. Since this happens on the backend, we shouldn't have a problem afaik.

paulovmr:
Hi @manstis, yes, it does. The preferences screen observes the `PreferencesCentralInitializationEvent` [here](https://github.com/paulovmr/uberfire/blob/b08acb938210a0c2ee985accdfe1d0c5bdde337b/uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/central/PreferencesCentralNavBarScreen.java#L45), and sets a custom scope resolution strategy, if one is passed. When the project admin page is built, we pass a custom scope resolution strategy supplier [here](https://github.com/paulovmr/guvnor/blob/1dea442342611fc1a878bf981f50ce88e9a46365/guvnor-project/guvnor-project-client/src/main/java/org/guvnor/common/services/project/client/ProjectEntryPoint.java#L56), that will be resolved when the user clicks in the item on the admin page, as we can see [here](https://github.com/paulovmr/uberfire/blob/b08acb938210a0c2ee985accdfe1d0c5bdde337b/uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/admin/page/AdminPageImpl.java#L136). So, in the case of the `GAVPreferences`edition in a project-level, its value will be saved only for that project.

manstis:
Thanks for the great explanations @paulovmr

paulovmr:
jenkins retest this

PR #389, opened 2016-11-18
Move injection to constructor for Bob Brodt

manstis:
@bbrodt would like this change to make his life easier.


manstis:
Hi @bbrodt if you can confirm this change fits your needs I'll merge it into master.

bbrodt:
Hi @manstis I see SessionInfo is no longer @Injected - is that a problem for existing code?

manstis:
@bbrodt Hi.. the constructor is marked with ```@Inject``` so CDI will inject all parameters; but now you can also re-use and pass explicit constructor parameters. I also tested the change.. and everything still works :-)

bbrodt:
Ah right, I didn't catch that. Awesome, ship it!

manstis:
Jenkins please retest this.

PR #391, opened 2016-11-23
Maven binary, add GAV properties and absolute path

cristianonicolai:
Add GAV attributes: artifactId, version and groupId to resolved MavenBinary.
 Resolve MavenBinary absolute path at the executor level, allowing reuse of both MavenBinaryImpl and MavenProjectBinaryImpl at Wildfly exec stage.


cristianonicolai:
@Salaboy @porcelli you would mind having a look?

nmirasch:
+1

PR #395, opened 2016-12-08
Add revapi plugin

MarianMacik:
Hi, @psiroky,

here is a revapi plugin for  guvnor-rest/guvnor-rest-backend module. It just checks REST endpoint called ProjectResource. There is again this bug in revapi, so 2 of 3 ignores are actually false-negatives.

Thanks,

Marian

### DIFF
--- a/guvnor-rest/guvnor-rest-backend/pom.xml
+++ b/guvnor-rest/guvnor-rest-backend/pom.xml
@@ -111,4 +111,13 @@
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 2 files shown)


manstis:
@psiroky Do you know much about ```revapi```? Can this be merged?

MarianMacik:
Hi, @manstis, we have already spoken about that with @psiroky, but he is currently not in office. Once he is there, we will merge it.

PR #398, opened 2016-12-21
GUVNOR-2813: Not possible to copy repository URL in Repository Editor

manstis:
See https://issues.jboss.org/browse/GUVNOR-2813

This PR adds a "notification" for when the URI is copied. It does not address the use of Flash which is up for debate (see the original [JIRA](https://issues.jboss.org/browse/RHBPMS-4123) ).


csadilek:
+1

PR #399, opened 2017-01-02
DROOLS-1374: Rename property version.wildfly.swarm to version.org.wil…

mbiarnes:
…dfly.swarm
needs to be merged with: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/339


csadilek:
@mbiarnes the PR build fails because the branch names don't line up. Your other branch is called wildflySwarm. Can you please adjust. We want green PR builds for everything that goes in. Thanks.

PR #409, opened 2017-01-28
Add constructors to allow use without requiring CDI

bbrodt:
https://issues.jboss.org/browse/GUVNOR-2872

See
org.kie.workbench.common.services.datamodel.backend.server.DataModelServiceConstructorTest
for unit test.

### DIFF
--- a/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationFactoryImpl.java
+++ b/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationFactoryImpl.java
@@ -27,9 +27,16 @@ import org.guvnor.structure.server.config.SecureConfigItem;
 
 public class ConfigurationFactoryImpl implements ConfigurationFactory {
 
-    @Inject
     private PasswordService secureService;
 
+    public ConfigurationFactoryImpl() {
+    }
+    
+    @Inject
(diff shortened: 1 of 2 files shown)


guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationFactoryImpl.java
```diff
@@ -30,6 +30,14 @@
     @Inject
```
manstis: As you're moving to constructor injection please remove the [instance variable injection](https://github.com/bbrodt/guvnor/blob/18f93d3d0b26bb9d8771e6614ce583b01e33a1ae/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationFactoryImpl.java#L42).

guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java
```diff
@@ -116,6 +116,27 @@
     @Named("systemFS")
     private FileSystem fs;
 
+    @Inject
```
manstis: Please remove the instance variable injection (as per ```ConfigurationFactoryImpl``` comment)

PR #410, opened 2017-01-30
Add constructors to allow use without requiring CDI

bbrodt:
https://issues.jboss.org/browse/GUVNOR-2872

See
org.kie.workbench.common.services.datamodel.backend.server.DataModelServiceConstructorTest
for unit test.

### DIFF
--- a/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java
+++ b/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java
@@ -74,32 +74,21 @@ public class ConfigurationServiceImpl implements ConfigurationService,
     // mainly for windows as *NIX is based on POSIX but escape always to keep it consistent
     private static final String INVALID_FILENAME_CHARS = "[\\,/,:,*,?,\",<,>,|]";
 
-    @Inject
-    @Named("system")
     private org.guvnor.structure.repositories.Repository systemRepository;
 
-    @Inject
     private ConfigGroupMarshaller marshaller;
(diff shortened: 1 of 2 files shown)


guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java
```diff
@@ -74,32 +74,23 @@
...
     private final Map<ConfigType, List<ConfigGroup>> configuration = new ConcurrentHashMap<ConfigType, List<ConfigGroup>>();
     private AtomicLong localLastModifiedValue = new AtomicLong( -1 );
 
-    @Inject
-    @Named("configIO")
     private IOService ioService;
 
     // monitor capabilities
-    @Inject
     @Repository
```
manstis: You'll need to move these qualifiers to the constructor; e.g. ```@OrgUnit Event<SystemRepositoryChangedEvent> orgUnitChangedEvent``` and delete them here.

guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java
```diff
@@ -74,32 +74,23 @@
...
     private Event<SystemRepositoryChangedEvent> repoChangedEvent;
-    @Inject
     @OrgUnit
```
manstis: You'll need to move these qualifiers to the constructor; e.g. ```@OrgUnit Event<SystemRepositoryChangedEvent> orgUnitChangedEvent``` and delete them here.

PR #413, opened 2017-02-06
Don't reference out-of-date playground repository in Guvnor ALA test.

mbarkley:
So I noticed a test failure on master was caused by failure to download GWT 2.8.0-rc1. Though the error was probably due to network flakiness, we shouldn't be relying on that GWT version anywhere and I think the source was from this test using an outdated playground repo.

This change points the test to my fork of the appformer-playground which is currently being maintained. I also noticed several other tests pointing to Mauricio's forks of "drools-workshop" that we might need to update in the future.


mbarkley:
Jenkins, retest this.

mbarkley:
Jenkins, retest this.

PR #416, opened 2017-02-17
PLANNER-736: Workbench: Prevent user from creating multiple @PlanningSolution data objects

mcimbora:
* Add copy constructor to ValidationMessage


mcimbora:
@manstis Wow, you're quick! Do you use any sort of bot thing? :-)

manstis:
@mcimbora We're all part of The Borg.

PR #419, opened 2017-03-02
[DOOLS-1452] fix guice-servlet dep after Maven 3.2.5 -> 3.3.9 upgrade

psiroky:


psiroky:
So the tests are failing, because they depend on https://github.com/Salaboy/drools-workshop/ (yikes!) and the projects there are not compatible with maven 3.3.9 (more precisely the docker-plugin).

I am not sure how it happened that the tests are depending on such "random" repo, but we need to fix that.

PR #421, opened 2017-03-08
PLANNER-704: Workbench: scoreHolder global can only be of org.optaplanner.core.api.score.holder.ScoreHolder type (or subclass)

mcimbora:
* Introduce GeneratedFileAttributesView

Useful to mark a file as "generated" and react accordingly on its processing. The file can be generated as a consequence of an action (e.g. creating a Planning Solution in OptaPlanner domain editor, etc.).

Part of:
https://github.com/droolsjbpm/guvnor/pull/421
https://github.com/droolsjbpm/drools-wb/pull/413
https://github.com/droolsjbpm/optaplanner-wb/pull/138

### DIFF
--- a/guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesView.java
+++ b/guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesView.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.common.services.backend.metadata.attribute;
+
+import java.util.Map;
+
+import org.uberfire.java.nio.IOException;
+import org.uberfire.java.nio.base.AbstractBasicFileAttributeView;
+import org.uberfire.java.nio.base.AbstractPath;
+import org.uberfire.java.nio.base.NeedsPreloadedAttrs;
+import org.uberfire.java.nio.file.attribute.BasicFileAttributeView;
+import org.uberfire.java.nio.file.attribute.BasicFileAttributes;
+import org.uberfire.java.nio.file.attribute.FileTime;
+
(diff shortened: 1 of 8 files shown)


guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesView.java
```diff
@@ -0,0 +1,102 @@
+package org.guvnor.common.services.backend.metadata.attribute;
```
manstis: License please

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedFileAttributes.java
```diff
@@ -0,0 +1,11 @@
+package org.guvnor.common.services.backend.metadata.attribute;
```
manstis: License please.

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedFileAttributes.java
```diff
@@ -0,0 +1,11 @@
...
+
+import org.uberfire.java.nio.file.attribute.BasicFileAttributes;
+
+/**
+ * Represents attributes of a file which has been generated.
+ */
+public interface GeneratedFileAttributes extends BasicFileAttributes {
```
manstis: I wonder whether we should extend ```BasicFileAttributes``` as (as far as I can tell) the "basic" view is still used in addition to the "generated" view - and hence you're duplicating many attributes. Is there not a _higher_ super interface in the hierarchy you can extend that provides the "View" definition without duplicating attributes?

guvnor-services/guvnor-services-backend/src/test/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedFileAttributesViewTest.java
```diff
@@ -0,0 +1,57 @@
+package org.guvnor.common.services.backend.metadata.attribute;
```
manstis: License please.

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedFileAttributes.java
mcimbora: @manstis Unfortunately UF currently imposes extending of `AbstractBasicFileAttributeView` https://github.com/uberfire/uberfire/blob/master/uberfire-io/src/main/java/org/uberfire/io/impl/IOServiceDotFileImpl.java#L188 in order for attributes to be loaded, therefore the hierarchy cannot be avoided (or at least I missed a way to do so!).

manstis: @mcimbora OK. Thanks for the explanation.

PR #425, opened 2017-03-22
Library issues fix

paulovmr:
GUVNOR-2952: [Library] Breadcrumb should be updated when the opened project is renamed
GUVNOR-2947: [Library] Fix create new asset dropdown overflow and order items alphabetically
GUVNOR-2946: [Library] Show business process, guided decision table, data object first at the empty project screen (and order the rest alphabetically).
GUVNOR-2951: [Library] Add option to create new folder/package in a project.

### DIFF
--- a/guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/AbstractProjectService.java
+++ b/guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/AbstractProjectService.java
@@ -53,7 +53,7 @@ import org.uberfire.java.nio.file.StandardDeleteOption;
 import org.uberfire.rpc.SessionInfo;
 import org.uberfire.security.authz.AuthorizationManager;
 
-import static org.guvnor.common.services.project.utils.ProjectResourcePaths.*;
+import static org.guvnor.common.services.project.utils.ProjectResourcePaths.POM_PATH;
 
 public abstract class AbstractProjectService<T extends Project>
         implements ProjectServiceCore<T>,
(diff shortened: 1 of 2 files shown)


paulovmr:
Part of an ensemble:

* https://github.com/kiegroup/guvnor/pull/425
* https://github.com/kiegroup/kie-wb-common/pull/739
* https://github.com/kiegroup/drools-wb/pull/428
* https://github.com/kiegroup/jbpm-designer/pull/559

guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/AbstractProjectService.java
```diff
@@ -157,198 +166,204 @@ public WorkingSetSettings loadWorkingSetConfig( final Path project ) {
...
                 throw e;
             } finally {
+                final Project newProject = resourceResolver.resolveProject(Paths.convert(newProjectPath));
+                renameProjectEvent.fire(new RenameProjectEvent(oldProject,
+                                                               newProject));
                 ioService.endBatch();
             }
-            final Project newProject = resourceResolver.resolveProject( Paths.convert( newProjectPath ) );
-            invalidateDMOCache.fire( new InvalidateDMOProjectCacheEvent( sessionInfo, oldProject, oldProjectDir ) );
-            renameProjectEvent.fire( new RenameProjectEvent( oldProject, newProject ) );
```
paulovmr: Real change is to move this line (and the `newProject` declaration) inside the `finally` block, so it is executed before the `ioService.endBatch();` call. This way, the `renameProjectEvent` is fired before the `deleteProjectEvent`, which is fired when the `pom.xml` is actually removed.

manstis:
@paulovmr Looks OK; however since the sequence of calls is important you could write a Unit Test verifying the order of invocations. See [InOrder](https://static.javadoc.io/org.mockito/mockito-core/2.2.9/org/mockito/InOrder.html).. just a thought.

paulovmr:
@manstis Done, thanks for (all) the reviews!

PR #431, opened 2017-04-03
GUVNOR-2971 : [GSS](6.4.x)Artifact repository upload functionality allows malicious code injection

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2971

### DIFF
--- a/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpPostHelper.java
+++ b/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpPostHelper.java
@@ -46,7 +46,7 @@ import static org.guvnor.m2repo.utils.FileNameUtilities.*;
 
 public class HttpPostHelper {
 
-    private static final Logger log = LoggerFactory.getLogger( HttpPostHelper.class );
+    private static final Logger log = LoggerFactory.getLogger(HttpPostHelper.class);
 
     @Inject
     private ExtendedM2RepoService m2RepoService;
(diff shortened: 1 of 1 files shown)


Rikkola:
@karreiro or who ever feels like they want to review, please take a look.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpPostHelper.java
```diff
@@ -208,9 +207,13 @@ private String uploadPom( final FormData formData ) throws IOException {
...
+                          pbe);
+                return UPLOAD_UNABLE_TO_PARSE_POM;
             } catch (MavenEmbedderException mee) {
-                return mee.getMessage();
+                log.error(mee.getMessage(),
+                          mee);
+                return UPLOAD_UNABLE_TO_PARSE_POM;
             }catch (Exception e) {
                 return UPLOAD_UNABLE_TO_PARSE_POM;
             }
```
psiroky: Is there a specific reason to _not_ log the general `Exception`? If not, the whole catch block could be simplified to just
```
catch (Exception ex) {
    log.error("Can no parse the provided POM", ex); // or some better log message
    return UPLOAD_UNABLE_TO_PARSE_POM;
}
```
which also brings different question: should we even catch these general `Exception`s? Shouldn't they be propaged and caught the the top-level `UncaughtExceptionHandler` (or how we call it)?

Rikkola: This is used by the file upload servlet so the UncaughtExceptionHandler might not pick it up, but true we should probably just capture any exception.

Rikkola:
@karreiro Updated with the exception change @psiroky mentioned. Functionality is still the same.

Rikkola:
jenkins please retest this

PR #442, opened 2017-04-17
GUVNOR-3009: [Library] Create a root screen on library (tiles) to navigate and create Organizational Units (Teams)

paulovmr:


paulovmr:
@Rikkola Hi Toni, if you could take a look please :)

PR #444, opened 2017-04-19
[7.0.x] GUVNOR-3009: [Library] Create a root screen on library (tiles) to navigate and create Organizational Units (Teams)

paulovmr:
Cherry picked from master (https://github.com/kiegroup/guvnor/pull/442).


paulovmr:
@Rikkola If you could take a look, just cherry picking my previous commits to 7.0.x.

PR #445, opened 2017-04-25
[6.5.x] RHBRMS-2792: [GSS] (6.4.z): WELD-001303 No active contexts for scope type javax.enterprise.context.RequestScoped in Tomcat

manstis:
See https://issues.jboss.org/browse/RHBRMS-2792

@tkobayas Hi, you could try this change locally on Tomcat.

It removes injection of "User" to a method parameter that can be correctly provided from the client.

If it solves the issue we can merge into our 6.5.x.

### DIFF
--- a/guvnor-inbox/guvnor-inbox-backend/src/main/java/org/guvnor/inbox/backend/server/InboxBackend.java
+++ b/guvnor-inbox/guvnor-inbox-backend/src/main/java/org/guvnor/inbox/backend/server/InboxBackend.java
@@ -17,23 +17,19 @@ package org.guvnor.inbox.backend.server;
 
 import java.util.List;
 
-/**
- * Created with IntelliJ IDEA.
- * Date: 9/30/13
- * Time: 4:06 PM
- * To change this template use File | Settings | File Templates.
- */
+import org.jboss.errai.security.shared.api.identity.User;
+
 public interface InboxBackend {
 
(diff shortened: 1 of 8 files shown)


manstis:
@etirelli @tkobayas has mentioned this fix will be provided as a "one of patch" to the affected customer.

Is it customary to also merge "one off patches" into our code-base?

manstis:
@tkobayas has confirmed this will be in BRMS 6.4.4; so until BRMS 6.4.3 is released this PR can wait...

manstis:
@karreiro Could you please review. 

There as such was no bug to fix; just change the way ```User``` instance was discovered. The old code injected it into a class ran on an unmanaged ```Thread``` on Tomcat which led to failures. The change herein removes the injection and instead passes it as a parameter. Existing tests cover the functionality.

manstis:
@karreiro Do you know which code formatter was used for 6.5.x? But; yes I agree I should have at least formatted all the files changed by the PR the same.. I'll update the PR.

karreiro:
@manstis I think we were using [this one](https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/6.5.x/ide-configuration). But, using the _new_ code style in all _new_ PRs sounds fair enough. It makes the cherry pick easier when we need to commit something on master and 6.5.x branch.

PR #446, opened 2017-04-26
Fixes for AppFormer

wmedvede:
- adds the ability of recurse into the super class hirerachy for finding the interfaces, instead of just querying the parent class.


wmedvede:
Hello @pefernan @porcelli  would you mind review this PR.

This adding is needed for the summit demo.
Thanks.

PR #452, opened 2017-05-19
Ignore Guvnor ALA failing tests in guvnor-ala-distribution

wmedvede:


wmedvede:
@manstis @porcelli @mbiarnes @krisv 
This is the PR for the two failing tests that are failng in guvnor-ala-distribution. As discussed by email the "best" option is to ignore then since nobody will use the guvnor-ala-distribution stuff at least until the 7.1 and let the 7.0.0.Final be.

psiroky:
If we are not going to use this for 7.0.0.Final wouldn't it be better to just delete the tests (or preferably the whole module)?

porcelli:
this distro will be needed very soon, so -1 to remove it. in the same.. have a jira associated to track the ignore is always good.

ederign:
agree with @porcelli . Can someone create a JIRA for this?

PR #453, opened 2017-05-19
Ignore Guvnor ALA failing tests in guvnor-ala-distribution

wmedvede:
@manstis @porcelli @krisv @mbiarnes 

This is the PR for the two failing tests that are failng in guvnor-ala-distribution. As discussed by email the "best" option is to ignore then since nobody will use the guvnor-ala-distribution stuff at least until the 7.1 and let the 7.0.0.Final be.

This is the 7.0.x counter part for the https://github.com/kiegroup/guvnor/pull/452


wmedvede:
Jenkins please retest this

manstis:
We could debate whether the test, class or module should be removed for 7.0.x; however simply to expedite a successful build of 7.0.x for 7.0.0.Final I'll merge this "as is".

psiroky:
If nothing else, the `@Ignore` annotation should at least contain a reason why the test was ignored (possibly with a link to JIRA).

Just my $0.02.

PR #456, opened 2017-05-24
[UF-506]: WorkspaceScoped beans

adrielparedes:
- Improvements on M2 Repositories. Now you can configure some of them through Preference API
- Some beans converted into WorkspaceScoped beans (https://github.com/AppFormer/uberfire/pull/718)
- It also exist a WorkspaceScoped Repository but it's disabled by default because other devs needs to be done before It can be used, but It is basically a repository that is going to be used by a Workspace to store artifacts, and those artifacts can only be seen my that workspace.

### DIFF
--- a/guvnor-m2repo-editor/guvnor-m2repo-editor-api/src/main/java/org/guvnor/m2repo/preferences/ArtifactRepositoryPreference.java
+++ b/guvnor-m2repo-editor/guvnor-m2repo-editor-api/src/main/java/org/guvnor/m2repo/preferences/ArtifactRepositoryPreference.java
@@ -24,20 +24,68 @@ import org.uberfire.preferences.shared.bean.BasePreference;
 @WorkbenchPreference(identifier = "ArtifactRepositoryPreference", bundleKey = "ArtifactRepositoryPreference.Label")
 public class ArtifactRepositoryPreference implements BasePreference<ArtifactRepositoryPreference> {
 
-    @Property(bundleKey = "ArtifactRepositoryPreference.DefaultM2RepoDir")
-    private String defaultM2RepoDir;
+    @Property(bundleKey = "ArtifactRepositoryPreference.GlobalM2RepoDir")
+    private String globalM2RepoDir;
+
+    @Property(bundleKey = "ArtifactRepositoryPreference.GlobalM2RepoDirEnabled")
+    private boolean globalM2RepoDirEnabled;
+
+    @Property(bundleKey = "ArtifactRepositoryPreference.WorkspaceM2RepoDir")
+    private String workspaceM2RepoDir;
+
+    @Property(bundleKey = "ArtifactRepositoryPreference.WorkspaceM2RepoDirEnabled")
+    private boolean workspaceM2RepoDirEnabled;
+
+    @Property(bundleKey = "ArtifactRepositoryPreference.DistributionManagementM2RepoDirEnabled")
+    private boolean distributionManagementM2RepoDirEnabled;
 
     @Override
     public ArtifactRepositoryPreference defaultValue(final ArtifactRepositoryPreference defaultValue) {
-        defaultValue.defaultM2RepoDir = "repositories/kie";
+        defaultValue.globalM2RepoDir = "repositories/kie/global";
+        defaultValue.workspaceM2RepoDir = "repositories/kie/workspaces";
+        defaultValue.globalM2RepoDirEnabled = true;
+        defaultValue.workspaceM2RepoDirEnabled = false;
+        defaultValue.distributionManagementM2RepoDirEnabled = true;
         return defaultValue;
     }
 
-    public String getDefaultM2RepoDir() {
-        return defaultM2RepoDir;
+    public String getGlobalM2RepoDir() {
+        return globalM2RepoDir;
     }
 
-    public void setDefaultM2RepoDir(final String defaultM2RepoDir) {
-        this.defaultM2RepoDir = defaultM2RepoDir.trim();
+    public void setGlobalM2RepoDir(final String globalM2RepoDir) {
+        this.globalM2RepoDir = globalM2RepoDir.trim();
+    }
+
+    public String getWorkspaceM2RepoDir() {
+        return workspaceM2RepoDir;
+    }
+
+    public void setWorkspaceM2RepoDir(String workspaceM2RepoDir) {
+        this.workspaceM2RepoDir = workspaceM2RepoDir;
+    }
+
+    public boolean isGlobalM2RepoDirEnabled() {
+        return globalM2RepoDirEnabled;
+    }
+
+    public void setGlobalM2RepoDirEnabled(boolean globalM2RepoDirEnabled) {
+        this.globalM2RepoDirEnabled = globalM2RepoDirEnabled;
+    }
+
+    public boolean isWorkspaceM2RepoDirEnabled() {
+        return workspaceM2RepoDirEnabled;
+    }
+
+    public void setWorkspaceM2RepoDirEnabled(boolean workspaceM2RepoDirEnabled) {
+        this.workspaceM2RepoDirEnabled = workspaceM2RepoDirEnabled;
+    }
+
+    public boolean isDistributionManagementM2RepoDirEnabled() {
+        return distributionManagementM2RepoDirEnabled;
+    }
+
+    public void setDistributionManagementM2RepoDirEnabled(boolean distributionManagementM2RepoDirEnabled) {
+        this.distributionManagementM2RepoDirEnabled = distributionManagementM2RepoDirEnabled;
     }
 }
--- a/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
+++ b/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
@@ -40,7 +40,6 @@ import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
-import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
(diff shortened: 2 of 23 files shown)


guvnor-m2repo-editor/guvnor-m2repo-editor-api/src/main/java/org/guvnor/m2repo/preferences/ArtifactRepositoryPreference.java
```diff
@@ -24,20 +24,68 @@
...
+    private boolean distributionManagementM2RepoDirEnabled;
 
     @Override
     public ArtifactRepositoryPreference defaultValue(final ArtifactRepositoryPreference defaultValue) {
-        defaultValue.defaultM2RepoDir = "repositories/kie";
+        defaultValue.globalM2RepoDir = "repositories/kie/global";
+        defaultValue.workspaceM2RepoDir = "repositories/kie/workspaces";
+        defaultValue.globalM2RepoDirEnabled = true;
+        defaultValue.workspaceM2RepoDirEnabled = false;
+        defaultValue.distributionManagementM2RepoDirEnabled = true;
```
manstis: What's the "Distribution Management M2 Repository directory"? A Project's ```pom.xml``` chain (from project and through all it's parents) can declare a ```<distributionManagement>``` section containing details of where the JAR should be deployed when built. This can presumably be different for each "workspace" (as different Users can have different effective poms); but I wonder what purpose this directory serves!? The configured distribution management section could point to a remote server over HTTP so a local directory seems irrelevant - unless I'm missing something obvious.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
```diff
@@ -51,78 +50,78 @@
...
         }
     }
 
-    public String getRepositoryURL() {
-        File file = new File(getM2RepositoryRootDir());
+    private ArtifactRepository getArtifactRepository(String name) {
+        return this.repositories
+                .stream()
+                .filter(artifactRepository -> artifactRepository.getName().equals(name))
+                .findFirst().orElseThrow(() -> new RuntimeException(String.format("Repository %s not found",
```
manstis: I love streams; don't get me wrong. But what's the significance of there being more than one artifact repository that matches the given name!? Is that guaranteed **never** to occur?

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
```diff
@@ -51,78 +50,78 @@
...
-        preferences.load();
         setM2Repos();
     }
 
     private void setM2Repos() {
-        final String M2_REPO_ROOT = FilenameUtils.separatorsToSystem(preferences.getDefaultM2RepoDir());
 
-        final String meReposDir = System.getProperty("org.guvnor.m2repo.dir");
+        this.repositories.addAll(this.artifactRepositoryFactory.getRepositories());
+        this.pomRepositories.addAll(this.artifactRepositoryFactory.getPomRepositories());
```
manstis: What are "pom repositories" vs "global" and "workspace"? I don't see them used anywhere.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
```diff
@@ -435,8 +434,9 @@ private void deployPom(final GAV gav,
                 .collect(Collectors.toList());
     }
 
-    public static String getPomText(final String path) {
-        final File file = new File(M2_REPO_DIR,
+    public String getPomText(final String path) {
```
manstis: You'll have to excuse me; I am probably a bit thick when it comes to understanding these changes. A User can be merrily editing a project in _their_ workspace. They've uploaded a new JAR to the Workbench's "Artifact Repository" and added it as a dependency to their project when they click "View pom" in the Project Editor's "Dependency grid" they'll get the pom from the "global" repository. This suggests the Workbench's "Artifact Repository" is considered "global". Is this correct?

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpGetHelper.java
```diff
@@ -46,35 +46,37 @@
...
+            response.sendError(HttpServletResponse.SC_NOT_FOUND);
             return;
         }
 
-        requestedFile = URLDecoder.decode( requestedFile,
-                                           "UTF-8" );
+        requestedFile = URLDecoder.decode(requestedFile,
+                                          "UTF-8");
+
+        String repositoryName = request.getParameter("repository");
```
manstis: Can you guarantee this?

When "browsing" the Workbench's repository in (for example, your web browser) and they enter a URL like ```http://server/kie-wb/maven2/org/cheese/porcelli/caps.jar``` an HTTP ```GET``` will be invoked which does not include a repository name. In addition if a User has added the workbench's Artifact Repository as a remote repository to another project's pom Maven will issue HTTP GETs when resolving dependencies from this remote. Those requests will not contain a repository name either.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpPutHelper.java
```diff
@@ -53,70 +53,67 @@ public void handle( final HttpServletRequest request,
...
+                response.sendError(HttpServletResponse.SC_NOT_FOUND);
                 return;
             }
 
-            pathInfo = URLDecoder.decode( pathInfo,
-                                          "UTF-8" );
+            pathInfo = URLDecoder.decode(pathInfo,
+                                         "UTF-8");
+
+            String repositoryName = request.getParameter("repository");
```
manstis: Is this guaranteed?

A ```mvn deploy``` from the CLI will issue a HTTP PUT but will not include a repository name.

guvnor-m2repo-editor/guvnor-m2repo-editor-api/src/main/java/org/guvnor/m2repo/preferences/ArtifactRepositoryPreference.java
adrielparedes: @manstis the problem here is that I chose a not so happy name =S. The only thing I wanted to do here is to replicate some behaviour that had this code before to "not to use distribution management section". The name is what is wrong, it should be distributionManagementEnabled, and it should enable or disable the posibility that GuvnorM2Repository to deploy an artifact in the distributionManagement setted into the pom.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
adrielparedes: @manstis they are used into deployArtifact and deployPom because for instance pom is not deployed through distributionManagement, so I had to do that difference.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
adrielparedes: @manstis this is because now you can add more than one repository, so you in when you try to do a HTTP GET to obtain an artifact you need to select the repository where the artifact is going to be search. If the name is not found it raises an exception. Anyway I'm going to change the exception for something more expressive.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpGetHelper.java
adrielparedes: @manstis yes, this is guaranteed, because the service that retrieves an artifact uses the default repository if it receives a null o empty string repository name. I tried not to change any default behavior.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpPutHelper.java
adrielparedes: @manstis the same than previous comment

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
adrielparedes: @manstis yes this is correct, but for now because I didn't want to change the default behavior since we need to do other more changes into workbench repositories before I can modify this. In this PR there are not workspace repositories enabled. But in a future you are going to be able to search pom text starting in your workspace repository, and if it is not found in the global one.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpGetHelper.java
manstis: @adrielparedes You may be able to guarantee this when we attempt to download a JAR from the ```M2RepoEditorPresenter``` screen; however I was meaning what if the workbench repository is added to a project's ```pom.xml``` and Maven (from the CLI) tries to download JARs; or someone is browsing the Workbench's Maven repository in a browser; etc. In these scenarios the ```HttpServletRequest``` is not controlled by *us* and hence may not have a ```repository``` parameter... or am I still missing something?

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java
ederign: at least please add i18n on the error msg.

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpGetHelper.java
```diff
@@ -84,187 +86,200 @@ public void handle( final HttpServletRequest request,
         long lastModified = file.lastModified();
         String eTag = fileName + "_" + length + "_" + lastModified;
 
-        String ifNoneMatch = request.getHeader( "If-None-Match" );
-        if ( ifNoneMatch != null && matches( ifNoneMatch, eTag ) ) {
-            response.setHeader( "ETag", eTag ); // Required in 304.
-            response.sendError( HttpServletResponse.SC_NOT_MODIFIED );
+        String ifNoneMatch = request.getHeader("If-None-Match");
```
ederign: This code deserves a refactoring someday! :-)

guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/helpers/HttpGetHelper.java
```diff
@@ -84,187 +86,200 @@ public void handle( final HttpServletRequest request,
...
+                response.setStatus(HttpServletResponse.SC_PARTIAL_CONTENT); // 206.
+
+                copyRange(input,
+                          output,
+                          r.start,
+                          r.length);
             } else {
-                response.setContentType( "multipart/byteranges; boundary=" + MULTIPART_BOUNDARY );
-                response.setStatus( HttpServletResponse.SC_PARTIAL_CONTENT ); // 206.
+                response.setContentType("multipart/byteranges; boundary=" + MULTIPART_BOUNDARY);
```
ederign: definitely deserves ;)

PR #461, opened 2017-06-05
Enable findbugs.failOnViolation

jhrcek:
Enable findbugs-maven-plugin to fail the build when serious bug has been identified.
At the same time fixing 2 issues identified by the plugin.

(The configuration of the plugin can be found in [droolsjbpm-build-bootstrap/pom.xml](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L1165-L1182) and has been already fine-tuned by QE to only fail the build in cases where it's almost certain there's a bug (as you can see in the issues fixed).

@manstis wdyt?


manstis:
@jhrcek seems a good idea (and has obviously proven useful already!)

PR #465, opened 2017-06-23
AF-547: VFS commit batch broken

mcimbora:
Introduce a wrapper around 'generated' value to make sure such value is not recognized as a Serializable in org.uberfire.java.nio.base.dotfiles.DotFileUtils.buildDotFile.

@ederign Would you mind taking a look? Thanks


mcimbora:
More changes required, will reopen when ready.

PR #473, opened 2017-07-07
AF-547: VFS commit batch broken

mcimbora:
Add 'generated' metadata attribute only for files that have actually been generated - this avoids generating dot files for non-generated files (which will likely be > 90 % cases).

Part of:
https://github.com/kiegroup/guvnor/pull/473
https://github.com/AppFormer/uberfire/pull/764

### DIFF
--- a/guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesUtil.java
+++ b/guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesUtil.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.common.services.backend.metadata.attribute;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class GeneratedAttributesUtil {
+
+    private GeneratedAttributesUtil() {
+    }
+
+    /**
+     * Remove attribute mappings specific to GeneratedAttributesView from the
+     * map of all available file attributes.
+     */
+    public static Map<String, Object> cleanup(final Map<String, Object> _attrs) {
+        final Map<String, Object> attrs = new HashMap<>(_attrs);
+
+        attrs.replace(GeneratedAttributesView.GENERATED_ATTRIBUTE_NAME, null);
+
+        return attrs;
+    }
+}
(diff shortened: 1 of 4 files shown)


guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesUtil.java
```diff
@@ -0,0 +1,34 @@
...
+
+public final class GeneratedAttributesUtil {
+
+    private GeneratedAttributesUtil() {
+    }
+
+    public static Map<String, Object> cleanup(final Map<String, Object> _attrs) {
+        final Map<String, Object> attrs = new HashMap<>(_attrs);
+
+        attrs.replace(GeneratedAttributesView.GENERATED_ATTRIBUTE_NAME, null);
```
ederign: this is made in order to clean up the current files right? If so, the right thing to do isn't convert false -> null and Boolean -> bool?

ederign: also, if this^  is true, I would like to have a comment on the class for future.

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesView.java
```diff
@@ -98,7 +98,13 @@ public Object fileKey() {
     private boolean extractGenerated() {
         final Map<String, Object> content = path.getAttrStorage().getContent();
 
-        return Boolean.parseBoolean( String.valueOf( content.get( GENERATED_ATTRIBUTE_NAME ) ) );
+        final Object generatedFileAttribute = content.get(GENERATED_ATTRIBUTE_NAME);
+
+        if (generatedFileAttribute instanceof Boolean) {
```
ederign: could you please clarify on which cases this is not a Boolean?

mcimbora: Just a defensive programming in case someone inserts a non-Boolean value for the key. Moreover the instanceof provides a null check for free.

guvnor-services/guvnor-services-backend/src/main/java/org/guvnor/common/services/backend/metadata/attribute/GeneratedAttributesUtil.java
mcimbora: @ederign Just followed the same pattern we use for other attribute types (e.g. https://github.com/kiegroup/guvnor/pull/473/files/683a7c522f14cfcef9869bcf0016d30a9c5eedab#diff-a24efd5c5afe6306ed0e2039004af29aR133), i.e. remove mappings for keys defined by given key. I'll add a comment with the explaination.

ederign: ok.

ederign: Thanks.

mcimbora:
Thanks for the review @ederign, PR updated.

mcimbora:
Jenkins retest this.

mcimbora:
Jenkins retest this.

mcimbora:
Jenkins retest this.

PR #476, opened 2017-07-19
GUVNOR-3128: Store Guvnor ALA state using VFS

wmedvede:


wmedvede:
@porcelli would you mind review this PR?
Thanks.

depends on:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/474

porcelli:
@jhrcek could you please prioritize the review of this PR?

guvnor-ala/guvnor-ala-spi/src/main/java/org/guvnor/ala/registry/impl/util/PageSortUtil.java
```diff
@@ -0,0 +1,80 @@
...
+                 sortOrder);
+            return result;
+        }
+        // if the values are less than  (pageSize * page) means that in the requested page there is no item
+        if (values.size() <= pageSize * page) {
+            return Collections.emptyList();
+        }
+        List<T> result = new ArrayList<>(pageSize);
+        List<T> allValues = new ArrayList<>(values);
+        for (int i = page * pageSize; i < (page * pageSize) + pageSize; i++) {
```
jhrcek: FIRST we're selecting subset of values and THEN we're sorting it? Will this behave correctly when the incoming values are not sorted according to `sort` parameter? I don't see such cases covered in tests..

guvnor-ala/guvnor-ala-ui/guvnor-ala-ui-backend/src/main/java/org/guvnor/ala/ui/backend/service/ProviderTypeServiceImpl.java
```diff
@@ -83,34 +92,40 @@ public ProviderType getProviderType(final ProviderTypeKey providerTypeKey) {
...
+                           false);
     }
 
     @Override
     public Map<ProviderType, ProviderTypeStatus> getProviderTypesStatus() {
         final Map<ProviderType, ProviderTypeStatus> result = new HashMap<>();
-
-        enabledProviders.values().forEach(providerType -> result.put(providerType,
-                                                                     ProviderTypeStatus.ENABLED));
+        readProviderTypeEnablements().entrySet().forEach(entry -> {
```
jhrcek: Take advantage of Java 8 stuff for to improve readabiligy
```
      readProviderTypeEnablements().forEach((providerType, isEnabled) -> 
                result.put(providerType, isEnabled ? ProviderTypeStatus.ENABLED : ProviderTypeStatus.DISABLED)
        );
```

guvnor-ala/guvnor-ala-docker-provider/src/main/java/org/guvnor/ala/docker/executor/DockerProviderConfigExecutor.java
```diff
@@ -29,23 +31,25 @@
...
 
     @Override
-    public Optional<DockerProvider> apply( final DockerProviderConfig dockerProviderConfig ) {
-        final DockerProviderImpl provider = new DockerProviderImpl( dockerProviderConfig.getName(), 
-                dockerProviderConfig.getHostIp(), dockerProviderConfig );
-        runtimeRegistry.registerProvider( provider );
-        return Optional.of( provider );
+    public Optional<DockerProvider> apply(final DockerProviderConfig dockerProviderConfig) {
+        final DockerProviderImpl provider = new DockerProviderImpl(
+                new DockerProviderConfigImpl(dockerProviderConfig.getName(),
```
wmedvede: I'm not the creator of this DockerXXXStuff in geneal, and the only change I did in this class applying formating.
But going back to your query, the parameter DockerProviderConfig dockerProviderConfig is just an interface, and DockerProviderConfig is not a ClonableConfig, so unless starting to do castings to see if it's ClonableConfig, etc, the cloning can't be directly applied.
Since this code was left in this way by the creator, etc, and it's not the priority right now, I'd prefer to not touch it.

guvnor-ala/guvnor-ala-docker-provider/src/main/java/org/guvnor/ala/docker/executor/DockerRuntimeExecExecutor.java
```diff
@@ -51,89 +53,104 @@
...
+        String shortId = id.substring(0,
+                                      12);
         String host = "";
         ContainerInfo containerInfo;
         try {
-            containerInfo = docker.getDockerClient( runtimeConfig.getProviderId() ).inspectContainer( id );
-            host = docker.getDockerClient( runtimeConfig.getProviderId() ).getHost();
-        } catch ( DockerException | InterruptedException ex ) {
-            throw new ProvisioningException( "Error Getting Docker Container info: " + id + "with error: " + ex.getMessage(), ex );
+            containerInfo = docker.getDockerClient(runtimeConfig.getProviderId()).inspectContainer(id);
```
wmedvede: Same comment above, this code was left this way by the creator.
But I agree, in this case I believe the first docker.getDockerClient is just for "throwing a DockerException" in cases of failures and the containerInfo is really never used.
I'll remove the containerInfo variable to make you happy but preserve the invocation.

guvnor-ala/guvnor-ala-registry-vfs/src/main/java/org/guvnor/ala/registry/vfs/VFSPipelineExecutorRegistry.java
```diff
@@ -0,0 +1,116 @@
...
+        registryHelper.deleteBatch(path);
+        super.deregister(taskId);
+    }
+
+    private void initializeRegistryRoot() {
+        try {
+            registryRoot = registryHelper.ensureDirectory(PIPELINE_EXECUTOR_REGISTRY_PATH);
+        } catch (Exception e) {
+            //uncommon error
+            logger.error("An error was produced during " + VFSPipelineExecutorRegistry.class.getName() +
```
wmedvede: Well, see that this fails typically in cases when the application "started" and the "VFS in general is broken".
This type or errors are let's say uncomon but also AFAIK there's no way to do let's say from a particular component a sort of application global status setup for setting the application in an application global error status. In general when  the VFS fails the application starts failing in different places, e.g. the Preferences can't be read, the initial indexing will fail, etc. So in this context of a failing VFS I believe it's ok to let this component just log that the it was not possible to initialize this directory.

wmedvede: I think the same I've tried to explain above. The failure in the initialization case is a sort of uncommon failure that occurs 99% in cases where the VFS wasn't initialized properly. And all the WF will be producing countless errors. So I believe that in the initialization case there's not much to do.

Assuming the initialization was ok, and the WB, etc., is working well, when things starts to be registered, I might happen the case where a particular registering can't be produced, and in this case it makes sense to catch, log the error, and re-throw the exception

guvnor-ala/guvnor-ala-registry-vfs/src/main/java/org/guvnor/ala/registry/vfs/VFSRegistryHelper.java
```diff
@@ -0,0 +1,239 @@
...
+
+    @PostConstruct
+    protected void init() {
+        try {
+            provisioningRootPath = fileSystem.getPath(PROVISIONING_BRANCH,
+                                                      PROVISIONING_PATH);
+            logger.debug("provisioningRootPath: " + provisioningRootPath.toUri());
+        } catch (Exception e) {
+            //uncommon error
+            logger.error("An error was produced during VFS registries directory initialization.",
```
wmedvede: Related to the comments above, the init() case will "never" happen, the readEntries will hopefully occur with low probabilty, but is different since e.g. if the conent of a registry entry was manually modified (I don't know who could be interested on doing this) we can have an unmarshalling error, or whatever. And this is why this case is also different to the init().

guvnor-ala/guvnor-ala-registry-vfs/src/main/java/org/guvnor/ala/registry/vfs/VFSRegistryHelper.java
```diff
@@ -0,0 +1,239 @@
...
+        return directoryPath;
+    }
+
+    /**
+     * Helper method calculating the MD5 digest for a String.
+     * @param content a String value for calculating the digest.
+     * @return returns the digest as a 32 character hex string.
+     */
+    public String md5Hex(String content) {
+        if (content == null) {
```
wmedvede: In general I don't like to use "values" to represent the semantic of null. I have seen largely people trying to represent null using values, specially in database areas, and then it's very funny to see database registers that has values like
"", "----", "NULL", "null", "nil" and even worst things.
But let's do it, and add a proper comment in the method javadoc.

guvnor-ala/guvnor-ala-registry-vfs/src/main/java/org/guvnor/ala/registry/vfs/VFSRegistryHelper.java
```diff
@@ -0,0 +1,239 @@
...
+            return null;
```
wmedvede: fixed

guvnor-ala/guvnor-ala-wildfly/guvnor-ala-wildfly-provider/src/main/java/org/guvnor/ala/wildfly/executor/WildflyRuntimeExecExecutor.java
```diff
@@ -28,61 +28,70 @@
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static org.guvnor.ala.util.RuntimeConfigHelper.buildRuntimeName;
 
-public class WildflyRuntimeExecExecutor<T extends WildflyRuntimeConfiguration> implements RuntimeBuilder<T, WildflyRuntime>,
-        RuntimeDestroyer,
-        FunctionConfigExecutor<T, WildflyRuntime> {
+public class WildflyRuntimeExecExecutor<T extends WildflyRuntimeConfiguration>
+        implements RuntimeBuilder<T, WildflyRuntime>,
+                   RuntimeDestroyer,
+                   FunctionConfigExecutor<T, WildflyRuntime> {
```
wmedvede: This code was left this way by the creator, and there are more places than 2) to change, but let's see if it's possible.

guvnor-ala/guvnor-ala-wildfly/guvnor-ala-wildfly-provider/src/main/java/org/guvnor/ala/wildfly/executor/WildflyRuntimeExecExecutor.java
```diff
@@ -28,61 +28,70 @@
...
 
-    private Optional<WildflyRuntime> create( final WildflyRuntimeConfiguration runtimeConfig ) throws ProvisioningException {
+    private Optional<WildflyRuntime> create(final WildflyRuntimeConfiguration runtimeConfig) throws ProvisioningException {
 
         String warPath = runtimeConfig.getWarPath();
-        final Optional<WildflyProvider> _wildflyProvider = runtimeRegistry.getProvider( runtimeConfig.getProviderId(), WildflyProvider.class );
+        final Optional<WildflyProvider> _wildflyProvider = runtimeRegistry.getProvider(runtimeConfig.getProviderId(),
+                                                                                       WildflyProvider.class);
 
         WildflyProvider wildflyProvider = _wildflyProvider.get();
```
wmedvede: see that there are a considerable amount of code that I in some way I've "inherited" plus millions of lines of code for implementing the UI, so yes, it's not possible to do a complete refactoring/covering of the inherited code at the same time I develop the UI, plus all the "missing" pieces for implementing the UI itself.
I'll look at this line, but, I believe we should focus on the LA. See that the LA will even not include the WF10 stuff.

guvnor-ala/guvnor-ala-wildfly/guvnor-ala-wildfly-provider/src/main/java/org/guvnor/ala/wildfly/executor/WildflyRuntimeExecExecutor.java
wmedvede: I'm not confident about this change when the the deadline for the LA is so tight. Let's evaluate this change for next iteration.

guvnor-ala/guvnor-ala-registry-vfs/src/main/java/org/guvnor/ala/registry/vfs/VFSBuildRegistry.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.guvnor.ala.registry.impl.BaseBuildRegistry;
+
+/**
+ * Builds are not persisted by now.
+ */
+@ApplicationScoped
+public class VFSBuildRegistry
```
wmedvede: 1) all the VFSXXXRegistry extends the BaseXXXRegistry counterpart.

2) the VFSXXXRegistries only override the required methods e.g. register(XXX) that is the place where we really want to write on the VFS. But they also want's to inherit certain behaviors from the parent class. Since It'll become inefficient to do all registry operations by reading all the elements form the VFS every time. 

3) I have some VFSXXXRegistries that still don't write in the VFS because we evaluated to write only some of them. But we might want to register more things in the VFS in next iteration.
Additionally since the registries are automatically plugable by CDI I don't want to package the guvnor-ala-registry-vfs and the guvnor-ala-registry-local at the same time. If I do this, I need to start adding @Specialized to some of the guvnor-ala-registry-vfs beans, and package two jars guvnor-ala-registry-vfs.jar and guvnor-ala-registry-local.jar in the generated webapp. And this can be confusing and bad looking.

guvnor-ala/guvnor-ala-registry-vfs/src/main/java/org/guvnor/ala/registry/vfs/VFSRegistryEntry.java
```diff
@@ -0,0 +1,86 @@
...
+package org.guvnor.ala.registry.vfs;
+
+/**
+ * Class for managing the contents of VFS based registries.
+ * This class is used as the bag for storing all the VFS registries entries. Changes in this class might affect
+ * de marshalling/unmarshalling of this registries.
+ */
+public class VFSRegistryEntry {
+
+    public static final String JSON = "json";
```
wmedvede: you made me think about something.... It'll take years before we support or even need multiple formats, so it has gone.

guvnor-ala/guvnor-ala-spi/src/main/java/org/guvnor/ala/pipeline/execution/impl/PipelineExecutorTaskImpl.java
```diff
@@ -133,10 +142,12 @@ protected Object clone() throws CloneNotSupportedException {
                                                                       executionId);
 
         clone.setPipelineStatus(this.getPipelineStatus());
-        stageStatus.entrySet().forEach(entry -> clone.setStageStatus(entry.getKey(), entry.getValue()));
-        stageError.entrySet().forEach(entry -> clone.setStageError(entry.getKey(), entry.getValue()));
+        stageStatus.entrySet().forEach(entry -> clone.setStageStatus(entry.getKey(),
```
wmedvede: yes.... and hopefully even this stageStatus.forEach(clone::setStageStatus); works...

guvnor-ala/guvnor-ala-spi/src/main/java/org/guvnor/ala/registry/impl/util/PageSortUtil.java
wmedvede: Pagination is not currently being used, i.e., we always return all the values.
And we tipically have 1 or 2 provider types, 3 or 4 providers, 1 or 2 pipelines, and some runtimes.
So all values are always returned.
This class was just moved from package, i.e. It's part of the "inherited" stuff, and so is the test.
So I expect that works fine. Again, if I have now to refactor/re-do/and provide missing test cases for all the stuff there, we'll never have something. So for this initial version that always returns all values I think it's fine.
After the LA this might be another area for improvements.

guvnor-ala/guvnor-ala-spi/src/main/java/org/guvnor/ala/marshalling/MarshallerRegistry.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.guvnor.ala.marshalling;
+
+/**
+ * General purpose contract for the registering of marshallers.
+ */
+public interface MarshallerRegistry {
+
+    void register(final Class clazz,
```
wmedvede: I tried it, and it's a good idea from the static typing perspective.
But It was not possible to make  work altogether  when the marshallers needs to be injected by using CDI in the code below.

@ApplicationScoped
public class VFSMarshallerRegistry
        extends BaseMarshallerRegistry {

    private static final Logger logger = LoggerFactory.getLogger(VFSMarshallerRegistry.class);

    private Instance<Marshaller> marshallerInstance;

    public VFSMarshallerRegistry() {
        //Empty constructor for Weld proxying
    }

    @Inject
    public VFSMarshallerRegistry(final @Any Instance<Marshaller> marshallerInstance) {
        this.marshallerInstance = marshallerInstance;
    }

    @PostConstruct
    public void init() {
        marshallerInstance.iterator().forEachRemaining(marshaller -> {
            logger.debug("Registering marshaller into VFSMarshallerRegistry, marshaller: " + marshaller);
            register(marshaller.getType(),
                     marshaller);
        });
    }
}

wmedvede:
@porcelli the modifications are in place. And the PR was rebased with master and updated.

jhrcek:
@wmedvede Thanks for implementing some of my proposals and explaining your design decisions. Please don't take any of my comments as personal criticism. I understand that the deadline is tight but I believe that investing a day to removing rough edges during PR review will save many days of troubleshooting/bug fixing in the future. The only suggestion that I still have would be to occasionally run IntelliJ's code analysis and check at least the "Probable bugs" section. The stuff reported there is usually relevant.

jhrcek:
@wmedvede This whole-project-reformatting https://github.com/kiegroup/guvnor/pull/478 was just merged. You might want to reformat your PR as well using the agreed upon KIE java conventions mentioned there and the merge conflicts should be gone..

wmedvede:
This PR needs to be closed since now must go to the 7.2.x branch

PR #478, opened 2017-07-21
Reformat whole repo using KIE Java Conventions

jhrcek:
Let's stop those monstrous mostly-formatting-changes PRs!
Formatting the entire repo using [KIE Java Conventions](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml) using the following settings in IntelliJ:
reformatsettings
No other changes done. I don't want to break other people's work, so I'm willing to keep reformatting / rebasing this PR until it gets merged ;-)


manstis:
@jhrcek Are you volunteering to do others too? ;-)

jhrcek:
@manstis sure, working on it as we speak ;-) Anything to remove those 100+ files diffs..

guvnor-ala/guvnor-ala-build-maven/src/main/java/org/guvnor/ala/build/maven/config/MavenBuildConfig.java
```diff
@@ -34,14 +34,14 @@
...
-        result.add( "package" );
+        result.add("package");
         return result;
     }
 
     default Properties getProperties() {
         final Properties result = new Properties();
-        result.setProperty("failIfNoTests", "false" );
+        result.setProperty("failIfNoTests",
+                           "false");
```
yurloc: @manstis @jhrcek @hasys Guys, this is a result of an error in the IDEA config introduced [here](https://github.com/kiegroup/droolsjbpm-build-bootstrap/commit/60e3f5bd9f99b925286a74df71de3cb2e35dab08#diff-9d663bfb476dd8a86f4f213da123bfa5R76). Method call arguments and methods declaration parameters should only be wrapped when the line is too long. It can be achieved by using the `Wrap if long` option in IDEA. Currently it is set to `Wrap always`, which is unnecessary and wrong in my opinion. That is not required nor recommended by [Java Code Conventions](http://www.oracle.com/technetwork/java/codeconventions-150003.pdf) mentioned by Michael in [PR#358](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/358#issuecomment-272396800). I also wonder about @ge0ffrey's opinion and if he intended to approve the `Wrap always` option or `Wrap if long`.

manstis: It's not an error; it is intentional. The common opinion was that splitting parameters onto individual lines made stepping through method invocations easier when debugging. To be honest I'm completely fed up with talking about code formatting. We followed @hasys proposal; we maintained what @mdproctor wanted (parameter splitting) and we got the best "buy in" we've probably had..

yurloc: @manstis Thank you for explanation. I'm glad that you mentioned the debugging because now I can imagine there's at least some advantage to choosing this style. It'll be easier for me to accept it if it's actually going to be implemented. 😄

I know there was some discussion but I didn't have the impression that there was a consensus even between those who appear to be in charge of the decision making. So I'm really wondering if this will really be applied globally, including optaplanner repository where the code style is pretty consistent and the change like this will be somewhat intrusive.

ge0ffrey: "To be honest I'm completely fed up with talking about code formatting." +1
Happy to see this PR being merged and seeing more consistency in guvnor/wb projects and it being in general (but not individually as pointed out by Jiri) more consistent with the original Sun Java conventions and the JBoss extension of those. **optaplanner-wb should also be formatted as such to be consistent with the rest of the wb projects.** That being said, I am not keen on reformatting the optaplanner core engine repository like this, as that one does follow the Sun Java conventions and JBoss extensions of those, until I have time to take a closer look to our dialect of that. Bikeshedding to be continued, I am sure :)

mcimbora: @ge0ffrey I already did that couple of weeks ago https://github.com/kiegroup/optaplanner-wb/pull/179

jhrcek: @mcimbora thank you, one less repo I have to argue about ;-)

PR #480, opened 2017-07-26
Fix compilation error after remove of JGitUtil

manstis:
Fix compilation error after remove of JGitUtil by https://github.com/AppFormer/uberfire/pull/779


manstis:
@wmedvede I ran the test locally but it took forever (never.. I got bored waiting) completed. 

Let's see what Jenkins makes of it.

manstis:
@mbiarnes @Rikkola FYI

manstis:
@wmedvede @porcelli It looks like changing to the "new" Command-based API leads to the test to fail.

I'll leave this with you two to resolve.

wmedvede:
@manstis I had even no idea that this change will be produced. 
@porcelli Would you mind take a look please? This change is only for 8.0.0 or also for the 7.2.0?

PR #488, opened 2017-08-02
GUVNOR-3258: UI Runtime Start/Stop/Remove operations

wmedvede:


wmedvede:
@porcelli would you mind review this PR?
Thanks!.

guvnor-ala/guvnor-ala-distribution/src/test/java/org/guvnor/ala/services/tests/RuntimeEndpointsTestIT.java
```diff
@@ -43,88 +43,57 @@
     @Ignore
     public void checkService() {
         Client client = ClientBuilder.newClient();
-        WebTarget target = client.target(APP_URL);
-        ResteasyWebTarget restEasyTarget = (ResteasyWebTarget) target;
-        RuntimeProvisioningService proxy = restEasyTarget.proxy(RuntimeProvisioningService.class);
+        WebTarget target = client.target( APP_URL );
```
jhrcek: Seems like you're not using [Kie Java conventions|https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml] for formatting :-(

porcelli:
Jenkins execute full downstream build

guvnor-ala/guvnor-ala-distribution/src/test/java/org/guvnor/ala/services/tests/RuntimeEndpointsTestIT.java
wmedvede: Good catch, well be fixed. On the other hand I can confirm I've been using the correct code styling template.

wmedvede:
this is now broken because of the opeshift provider stuff in the middle, will have to be rebased against 7.2.x and updated

wmedvede:
@porcelli, @psiroky  this is another example of the problem I mentioned before.
An hour ago or so (approximately) this PR was green/green.
Now it "automatically" started a new build, no changes where done, and not r-e-tst command was executed.
The question is  Why is this new build started?

psiroky:
@wmedvede looking at the jobs, they have been executed only once today, so they definitely did not run an hour ago or so. What likely happened is that the jobs were is Jenkins queue for some time and the green/green you saw was actually quite old, before your recent changes.

guvnor-ala/guvnor-ala-provisioning-pipelines/guvnor-ala-provisioning-pipelines.iml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module version="4">
```
porcelli: file should be removed and add to the .gitignore

guvnor-ala/guvnor-ala-services-api/src/main/java/org/guvnor/ala/services/api/PipelineService.java
```diff
@@ -120,15 +120,36 @@ String newPipeline(@NotNull PipelineConfig config,
...
+
+    /**
+     * Deletes a pipeline execution
+     * @param executionId A pipeline execution id to delete. The pipeline execution id is typically returned by
+     * the runPipeline method.
+     * @throws BusinessException
+     */
+    @POST
+    @Path("execution/{executionId}/delete")
+    void deletePipelineExecution(@PathParam("executionId") String executionId) throws BusinessException;
```
porcelli: this should use the proper delete verb (https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html) and remove the "delete" from path

guvnor-ala/guvnor-ala-services-api/src/main/java/org/guvnor/ala/services/api/RuntimeProvisioningService.java
```diff
@@ -109,16 +109,28 @@ ProviderList getProviders(@QueryParam("page") @DefaultValue("0") Integer page,
...
+     * Destroy an existing Runtime by removing it from the runtime registries independently of the connectivity with
+     * the external provider. e.g. if it was not possible to connect an external WF where the runtime
+     * is running.
+     * @param runtimeId the identifier of the runtime to destroy
+     * @throw BusinessException in case of an internal exception
+     */
+    @DELETE
+    @Consumes(value = APPLICATION_JSON)
+    @Path("runtimes/{id}/forceDestroy")
+    void forceDestroyRuntime(@PathParam(value = "id") String runtimeId) throws BusinessException;
```
porcelli: same.. i think you could have the destroy and destroy as just delete operation... and we could get a boolean as parameter to force or not.

wmedvede:
@porcelli the lastly suggested modifications are in place

PR #498, opened 2017-08-17
GUVNOR-3025: OpenShift S2I (Source to Image) Integration

errantepiphany:
GUVNOR-3025: OpenShift S2I (Source to Image) Integration
https://issues.jboss.org/browse/GUVNOR-3025


errantepiphany:
These go together:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/516
https://github.com/kiegroup/droolsjbpm-integration/pull/1136
https://github.com/kiegroup/guvnor/pull/498

guvnor-ala/guvnor-ala-openshift/guvnor-ala-openshift-provider/src/main/java/org/guvnor/ala/openshift/access/OpenShiftTemplate.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Gets an openshift template by uri.
+     * @param client the client
+     * @param prjName the project/namespace
+     * @param templateUri the location of the template
+     * @return the template
+     * @throws OpenShiftClientException
+     */
+    public static OpenShiftTemplate fromUri(OpenShiftClient client, String prjName, String templateUri) {
```
wmedvede: Hi @errantepiphany  , quick query regarding the template parsing.
Why do we need a OpenShiftClient configuration for "parsing" the template?.
Is this client configuration mandatory?
I'm thinking about the following use case we talked about with @porcelli 
1) the new deployment wizard starts
2) the uri to the template is taken from "some place", e.g., a system property as Alex suggested for this initial iteration.
3) we get the OpenShiftTemplate by using the uri
4) finally we use the OpenShiftTemplate for building the UI
when we do 1,2,3,4 we should be able to "parse" the template independently of the openshift client

errantepiphany: Hi @wmedvede , it's because the client.getDelegate() is the thing that does the parsing of the template. However, you can use an "unconfigured" client to pass in. To see an example of this, see OpenShiftAccessTest.testTemplateParams().

wmedvede: yes, I saw quickly the OpenShiftAccessInterfaceImpl().newOpenShiftClient() that creates the OpenShiftClient and for a moment I thought a client created this way took some system settings. 
But seems like not, ok, thanks.

wmedvede:
@porcelli I did an overall look into this and seems good to me. If you can take a look it'll be great since this is what we need now to try to complete the integration.
Thanks.

PR #503, opened 2017-08-23
fix docker registry location in streams

errantepiphany:
fix docker registry location in streams
https://issues.jboss.org/browse/GUVNOR-3025


wmedvede:
Don't know why this PR is not yet being built. Probably a jenkings job is queued for it.
Let's give it some more time, if nothing happens let's force the build

psiroky:
Jenkins and github are having bad day. I am trying to find and fix the root cause. Jenkins will be restarted in the next or so and I hoping it will temporarily help.

wmedvede:
ok, thanks

PR #509, opened 2017-09-06
JBPM-6399 - Execution Server - Process config is not aligned with dep…

mswiderski:
…loyment descriptor

do not merge, pull requests intended to verify the approach before fine tuning and added test coverage. Explanation will follow in a comment...

### DIFF
--- a/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/builder/model/BuildResults.java
+++ b/guvnor-project/guvnor-project-api/src/main/java/org/guvnor/common/services/project/builder/model/BuildResults.java
@@ -18,7 +18,9 @@ package org.guvnor.common.services.project.builder.model;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.guvnor.common.services.project.model.GAV;
 import org.guvnor.common.services.shared.message.Level;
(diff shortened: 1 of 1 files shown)


mswiderski:
First things first, all PRs required for this change are:
https://github.com/kiegroup/guvnor/pull/509
https://github.com/kiegroup/kie-wb-common/pull/1135
https://github.com/kiegroup/jbpm-wb/pull/865

**Problem**
the problems is described in jira - but in short - when project has deployment descriptor and it does declare another than default runtime strategy (other than singleton) then this strategy is not reflected in process config on automatically created container spec.

**Solution**
make use of PostBuildHandler to read deployment descriptor from the project and set is as parameter in BuildResults. Then that value is taken by the BuildExecutor and used where creating ContainerSpec. Thus making all inline and preventing from various unexpected runtime issues.

**Things to look at**
I extended BuildResults to allow to pass simple parameters between parties involved in post build operations. The main reason was to reuse already prepared mechanism - PostBuildHandler - and avoid any additional server round trips - BuildExecutor that does create container spec is client side thingy.

Another thing is that parameters of BuildResults is used by the builder too, to pass the root path of the project to handlers. We could change the method signature or PostBuildHandler but don't know how much impact would that have...

So Guvor PR extends BuildResults, ie-wb-common PR makes use of the parameters of the BuildResults, and jbpm-wb PR implements PostBuildHandler to read deployment descriptor.


@manstis @karreiro @cristianonicolai could you please have a look at this and share your comments?

mswiderski:
Thanks @tsurdilo @manstis 

@manstis as soon as eclipse formatter will be merged I'll be more than happy to reformat but till then I will skip any formatting ralated comments, really have enough of that

manstis:
Awwwwww.. how can you say such a thing!

Perhaps you'd like to help @hasys with the Eclipse formatter? ;-)

mswiderski:
Isn't there one ready? And just to be merged... it just takes too much of time to constantly have the same discussion

PR #522, opened 2017-10-04
Windows test stabilization

mackaypeter:
Some minor fixes for Windows-specific test failures.

### DIFF
--- a/guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/MavenDependencyConfigExecutorTest.java
+++ b/guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/MavenDependencyConfigExecutorTest.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 
 import org.apache.commons.io.FileUtils;
@@ -70,6 +71,7 @@ public class MavenDependencyConfigExecutorTest {
         final String groupId = "org.guvnor.ala";
         final String artifactId = "maven-ala-artifact-test";
         final String version = "1";
+        final String artifactPath = "/org/guvnor/ala/maven-ala-artifact-test/1/maven-ala-artifact-test-1.pom";
 
         final String oldSettingsXmlPath = System.getProperty(CUSTOM_SETTINGS_PROPERTY);
         try {
@@ -116,14 +118,14 @@ public class MavenDependencyConfigExecutorTest {
                          binary.getArtifactId());
             assertEquals(version,
                          binary.getVersion());
-            assertEquals(m2Folder + "/org/guvnor/ala/maven-ala-artifact-test/1/maven-ala-artifact-test-1.pom",
+            assertEquals(Paths.get(m2Folder.getPath(), artifactPath).toString(),
                          binary.getPath().toString());
         } finally {
             if (oldSettingsXmlPath == null) {
                 System.clearProperty(CUSTOM_SETTINGS_PROPERTY);
             } else {
                 System.setProperty(CUSTOM_SETTINGS_PROPERTY,
-                                   oldSettingsXmlPath);
+                        oldSettingsXmlPath);
             }
             MavenSettings.reinitSettings();
         }
--- a/guvnor-ala/guvnor-ala-services-rest/src/test/java/org/guvnor/ala/services/rest/tests/RestPipelineImplTest.java
+++ b/guvnor-ala/guvnor-ala-services-rest/src/test/java/org/guvnor/ala/services/rest/tests/RestPipelineImplTest.java
@@ -10,6 +10,7 @@ import javax.inject.Inject;
 
 import com.spotify.docker.client.DockerException;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.SystemUtils;
 import org.guvnor.ala.build.Project;
 import org.guvnor.ala.build.maven.config.MavenBuildConfig;
 import org.guvnor.ala.build.maven.config.MavenProjectConfig;
@@ -84,7 +85,9 @@ import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.junit.After;
+import org.junit.Assume;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.uberfire.commons.lifecycle.Disposable;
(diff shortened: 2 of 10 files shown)


guvnor-ala/guvnor-ala-wildfly/guvnor-ala-wildfly-provider/src/test/java/org/guvnor/ala/wildfly/executor/tests/requirement/RequiresNotWindows.java
```diff
@@ -0,0 +1,15 @@
+package org.guvnor.ala.wildfly.executor.tests.requirement;
```
manstis: License please

guvnor-ala/guvnor-ala-wildfly/guvnor-ala-wildfly-provider/src/test/java/org/guvnor/ala/wildfly/executor/tests/requirement/NotWindowsRequirement.java
```diff
@@ -0,0 +1,17 @@
+package org.guvnor.ala.wildfly.executor.tests.requirement;
```
manstis: License please

guvnor-ala/guvnor-ala-wildfly/guvnor-ala-wildfly-provider/src/test/java/org/guvnor/ala/wildfly/executor/tests/requirement/NotWindowsRequirement.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.arquillian.cube.spi.requirement.Requirement;
+import org.arquillian.cube.spi.requirement.UnsatisfiedRequirementException;
+
+import java.util.Locale;
+
+public class NotWindowsRequirement implements Requirement<RequiresNotWindows> {
```
triceo: I was wondering - did we actually check it on something else but Linux and Windows? If not, wouldn't it be safer to just make this `Requirement` be `OnlyLinuxRequirement`?

guvnor-ala/guvnor-ala-services-rest/src/test/java/org/guvnor/ala/services/rest/tests/RestRuntimeProvisioningImplTest.java
```diff
@@ -132,6 +136,11 @@ public static Archive createDeployment() throws Exception {
         return deployment;
     }
 
+    @BeforeClass
+    public static void beforeClass() {
+        Assume.assumeFalse(System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("windows"));
```
triceo: Any chance we could share the conditional check with the `Requirement` below? Perhaps some utility method, such as `isWindows()`?

guvnor-ala/guvnor-ala-services-rest/src/test/java/org/guvnor/ala/services/rest/tests/RestPipelineImplTest.java
```diff
@@ -185,6 +188,11 @@ public static Archive createDeployment() throws Exception {
         return deployment;
     }
 
+    @BeforeClass
+    public static void beforeClass() {
+        Assume.assumeFalse(System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("windows"));
```
triceo: Considering that this `beforeClass()` is repeated in 3 different tests, wouldn't it make sense to create a new parent class for these tests and put the method there? Let's not duplicate code.

guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/MavenDependencyConfigExecutorTest.java
```diff
@@ -116,14 +118,14 @@ public void testMavenDependencyAPI() throws Exception {
                          binary.getArtifactId());
             assertEquals(version,
                          binary.getVersion());
-            assertEquals(m2Folder + "/org/guvnor/ala/maven-ala-artifact-test/1/maven-ala-artifact-test-1.pom",
+            assertEquals(Paths.get(m2Folder.getPath(), artifactPath).toString(),
```
triceo: I think it should be possible to compare `Path` directly, without the `toString()` - no?

guvnor-ala/guvnor-ala-services-rest/src/test/java/org/guvnor/ala/services/rest/tests/RestPipelineImplTest.java
baldimir: I don't think this would be good, because the classes are in different packages that are probably unrelated. However I am not against having a util class with a method that checks os type. E.g. like isWindows(). This method can then be used in the assumes here. Later, we can search for this method references so we will always know where we have put these assumes.

guvnor-ala/guvnor-ala-build-maven/src/test/java/org/guvnor/ala/build/maven/executor/MavenDependencyConfigExecutorTest.java
mackaypeter: The two Paths point at the same file, but are not actually equal as Path objects.

triceo: Ah, my bad. I misunderstood what Path does.

guvnor-ala/guvnor-ala-services-rest/src/test/java/org/guvnor/ala/services/rest/tests/RestPipelineImplTest.java
baldimir: Maybe this [1] can be also used when commons-lang is in the dependencies list. 

[1] http://grepcode.com/file/repo1.maven.org/maven2/commons-lang/commons-lang/2.6/org/apache/commons/lang/SystemUtils.java#SystemUtils.0IS_OS_WINDOWS

guvnor-ala/guvnor-ala-spi/src/test/java/org/guvnor/ala/marshalling/BaseMarshallerTest.java
```diff
@@ -80,8 +82,14 @@ public void testGetClass() {
     @Test
     public void testMarshall() throws Exception {
         String marshalledValue = marshaller.marshal(getValue());
-        assertEquals(getMarshallerOutput(),
-                     marshalledValue);
+        String marshallerOutput = getMarshallerOutput();
+        ObjectMapper objectMapper = new ObjectMapper();
+        Map<String, Object> marshalledValueMap =
```
wmedvede: good trick

mackaypeter:
Thanks for the reviews, the requested changes are in.

manstis:
@wmedvede Is this OK with you? Could you please approve?

PR #527, opened 2017-10-11
JBPM-6337: Windows don't remove last File separator resolving reposit…

nmirasch:
…ory dir causing bad gav resolution
Changed the way that getM2RepositoryDiv removes the last File.separator to make it work properly on windows OS


nmirasch:
@adrielparedes could you take a look? thanks!

nmirasch:
@manstis @adrielparedes I'm going to close this pr, because I think the fix has been included at https://github.com/kiegroup/guvnor/commit/91c48288de9e2c3803522c5b8926d242ffeb83f9. I'm going to verify if it solves the issue that were reported aboud the gav resolution on windows.
Thanks both!

PR #529, opened 2017-10-20
GUVNOR-3521: DRL Editor: Declared types are ignored by the Fact Types list

manstis:
See https://issues.jboss.org/browse/GUVNOR-3521


manstis:
Part of an ensemble:
- https://github.com/kiegroup/guvnor/pull/529
- https://github.com/kiegroup/kie-wb-common/pull/1256

Rikkola:
The old tests should cover this, but of course best to check there are old tests.

manstis:
@jomarko Added new commit with missing Unit Tests for existing classes.

